PR #1, opened 2019-04-25
Intro page - Update references from iOS to mobile_app, tweak notify group example

SeanPM5:
Updated references from iOS to mobile_app, also tweaked the notify group example to a slightly easier to understand one imo, that also shows you're not limited to sending to just phones but can do tablets too.

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -2,7 +2,7 @@
 title: "Introduction"
 ---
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @SeanPM5, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#1).

Thanks!

PR #94, opened 2019-07-05
Improve Critical Notifications last sentence

AlecRust:
This is confusing:

> Note that have expanded the `sound` attribute to include the `critical: 1` flag and `volume: 1.0` to set the volume at 100 %.

Changed to:

> Note the example expands the `sound` attribute to include the `critical: 1` flag, and `volume: 1.0` to set the volume to 100%.

### DIFF
--- a/docs/notifications/critical.md
+++ b/docs/notifications/critical.md
@@ -29,4 +29,4 @@ automations:
 
 ```
 
-If you have previously read the [sounds documentation](sounds.md) this syntax should be mostly familiar. Note the example expands the `sound` attribute to include the `critical: 1` flag, and `volume: 1.0` to set the volume to 100%.
+If you have previously read the [sounds documentation](sounds.md) this syntax should be mostly familiar. Note the example expands the `sound` attribute to include the `critical: 1` flag, and `volume: 1.0` to set the volume to 100 %.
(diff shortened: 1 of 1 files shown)


docs/notifications/critical.md
```diff
@@ -29,4 +29,4 @@ automations:
 
 ```
 
-If you have previously read the [sounds documentation](sounds.md), this syntax should be mostly familiar. Note that have expanded the `sound` attribute to include the `critical: 1` flag and `volume: 1.0` to set the volume at 100 %.
+If you have previously read the [sounds documentation](sounds.md) this syntax should be mostly familiar. Note the example expands the `sound` attribute to include the `critical: 1` flag, and `volume: 1.0` to set the volume to 100%.
```
TomBrien: Great catch, thank you. I would prefer to keep the SI styling of 100 % rather than going down a particular national style guide. Happy to let the Oxford comma slip away though. This will need to be incorporated into the versioned docs but I'm happy to do that if you're ok with the small suggested change I will merge. Cheers again. 
```suggestion
If you have previously read the [sounds documentation](sounds.md) this syntax should be mostly familiar. Note the example expands the `sound` attribute to include the `critical: 1` flag, and `volume: 1.0` to set the volume to 100 %.
```

PR #140, opened 2019-12-25
Add missing URL Handler document

olbjan:
This was missing in 2.0.0 docs.


TomBrien:
Unless there are changes to the `version-1.0.0` doc there is no need to duplicate under `version-2.0.0` as it just falls back to using 1.0.0

PR #145, opened 2020-02-10
Begin adding Android examples and troubleshooting

dshokouhi:
There is a lot to do in terms of generalizing the docs more but I wanted to get the examples and basic troubleshooting steps in so we have something to build upon.  I hope what I tried to do makes sense as I was trying to keep things generalized.  Android does not support much yet so I didn't feel the need to touch on those areas until we have something.  For now users should have enough info to work with what is supported.

Definitely open to any suggestions as this my first pass at it.  Also had some issues previewing the changes with the icons so I hope it looks nice lol

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -12,7 +12,7 @@ Some useful examples of actionable notifications:
 
 ![Actionable notifications allow the user to send a command back to Home Assistant.](assets/ios/actions.png)
 
-## Overview of how actionable notifications work
+## Overview of how actionable notifications work on iOS
 
 In advance of sending a notification:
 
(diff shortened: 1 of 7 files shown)


dshokouhi:
@JBassett FYI, I am sure you can think of more that needs to be added but wanted to get something out there for now :)

JBassett:
Totally agree with getting something out there.  Documentation for this is well over due and not my strong suit....  Thanks for picking up my slack ;-)

JBassett:
Also side note, the preview site from netlify doesn't seem to reflect the changes?

dshokouhi:
Yea I noticed that myself.  Not sure why its not showing the updates on the preview site :/

TomBrien:
If edits are only made to `/docs` you need to change the version to `next`. Are you running docusaurus locally or just editing via the web?

TomBrien:
Looks good! Your preview is here! I’ll port those changes back to the landing version of the docs when I’m up properly

dshokouhi:
I did the edits from the Edit link on the docs pages.  I did not setup a development environment for this.

PR #147, opened 2020-02-10
Generalizing - Round 2

dshokouhi:
Continuing along from #145  I have gone through the remaining pages and added icons and more android notes.  Apologize for not having the development environment setup.  I will look into doing that later today/tomorrow so the next update won't require more fiddling to get it to work :)

The location section can use some more explanation as to how location updates from Android works but I think for now we have the important tid bits around permissions added.

The action section may need to be changed to properly reflect the example I provided but for now I think it should help users understand what they can do.

There is still more to add but at this point we can probably start focusing on more specifics like adding screenshots and walking users through certain steps.

These updates were done using the Edit links from the UI so I think I also need to copy them over to versioning?  Not entirely sure how that works just yet :)

### DIFF
--- a/docs/core/actions.md
+++ b/docs/core/actions.md
@@ -2,27 +2,27 @@
 title: "Actions"
 ---
 
-Actions is a generic system that allows you to easily integrate the Home Assistant automations system into multiple areas of iOS and [Apple Watch](../integrations/watch/index.md).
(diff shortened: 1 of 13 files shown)


TomBrien:
No worries at all! I’m going to look at migrating the site this week so it should be simpler soon (fingers crossed). I’ll port this back when I get home.

PR #152, opened 2020-02-11
Migrate to docusaurus V2

### LINKED ISSUE #151: Discussion: remove versioning and upgrade to Docusaurus v2
So as discussed on Discord a little. I think versioning is causing us a bit of pain. Main issue is that when someone uses the edit function these edits are made against the `next` version docs and require moving back to the `2.0.0` which are the landing version. 

I've built a demo version of the docs site [here](https://ha-appv2.tombrien.io/) from scratch and used Docusaurus 2 to do so. I ported over the conent but this is a couple of PRs out of date now but it is pretty much just a copy and paste the files job to update. Little summary

### Cons
-   Docusaurus 2 is at a late-stage alpha at the moment, although it is pretty well documented and seemed fine to me
-   No translations at the moment, these are coming in Docusaurus 2 soon I believe, and I don't think they were working on the old site to be honest
-   I found Docusaurus 2 requires slightly stricter markdown. I.e. underscores required escaping or putting in code ticks (I see no case where the ever should not be!)
-   We lose the versioned docs (1.x had little in it). The can be enabled down the line if needed again if we have a version with major changes.

### Pros
-   I was able to build a fresh version of the site with no leftovers from versioning
-   Docusaurus 2 has more useful behind the scenes features and more coming online
-   Edits done online will be in the landing version when merged with no need to port back to versioned docs
-   Pretty dark mode!
-   We lose the versioned docs!
### Notes
I haven't enabled the search bar yet for the proof of concept page but this should be trivial on the deployment

TomBrien:
Along with migrating to Docusaurus V2 this also removes the versioning (a good thing!) and translations (a bad thing ☹but they weren't active anyway). This has been discussed a little in #151 and will close #151.

Requires the netlify build to be changed to simply `yarn build ` and the publish directory should be `build` @robbiet480 not sure if when you implement this will Netlify attempt to rebuild the PR preview with the change?


TomBrien:
FYI these are the build settings from the demo site I put together
image

PR #166, opened 2020-02-13
Convert feature comparison table to html

TomBrien:
Convert feature comparison  table to html to allow column alignment while keeping the section headers and row shading (mid table headers not supported by markdown)


dshokouhi:
This looks much better, thank you!!

PR #173, opened 2020-02-14
Add widgets to comparison chart and fix links

dshokouhi:
Tiny bit of title restructuring so the link in the chart works better.  Also the logos look much better with the title font size as well :)


dshokouhi:
While testing this out it looks like the links in the chart arent working....looking into that as well

dshokouhi:
Hmm links are still failing but they work locally in `yarn start` :/

dshokouhi:
Now all the links are fixed :)

PR #189, opened 2020-02-21
Some more generalizing, tweaks and additions

dshokouhi:
- Made a few of the troubleshooting steps a bit more general.
- Included more iOS and Android specific areas.
- Added another Android troubleshooting step.
- Other tweaks.

I noticed that the section about sensor names needed to be adjusted.  There is a change in 0.106 that will make sure newly added sensors get the device name added so they should be easier to find.  Not sure how we feel about calling out a specific version of HA core but the current information was not correct.  The wording here will need to change slightly once the new version is out of beta.

### DIFF
--- a/docs/troubleshooting/setup.md
+++ b/docs/troubleshooting/setup.md
@@ -10,7 +10,7 @@ Once you have [set up](/getting_started/index.md) the Companion app you will nee
 
 ![iOS](/assets/apple.svg) If you don't see the service call on iOS, check the notification settings within the app (swipe right to bring up the sidebar, the tap "App Configuration", then "Notifications"). If the "Push ID" box is empty, tap the Reset button below it.
 
-![android](/assets/android.svg) If you still don't see the service call on Android follow the steps to [start fresh](#starting-fresh-with-the-android-app)
+![android](/assets/android.svg) If you still don't see the service call on Android follow the steps to [start fresh](#starting-fresh-with-the-android-app).
 
 #### I have a `notify.mobile_app_<Device_ID>` service but don't receive notifications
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/setup.md
```diff
@@ -67,16 +69,16 @@ At times you may need to start fresh with the Android app as a new feature may n
...
-2.  Turn off battery optimizations for the app
-3.  Ensure that the location toggles are enabled in App Configuration in the Android app
-4.  Turn on unrestricted data for the Android app
+1.  Ensure the app has location permissions granted, all the time.
+2.  Turn off battery optimizations for the app.
+3.  Ensure that the location toggles are enabled in App Configuration in the Android app.
+4.  Turn on unrestricted data for the Android app.
 
 #### Using a self-signed certificate leads to a blank page in Android
 If you are using a self-signed certificate on Android then you may get stuck at a blank screen after entering and/or selecting your Home Assistant instance. In order to correct this issue you will need to make sure the URL is valid and that you import the certificate into Androids Trusted Certificates. Steps to perform this can be found [here](https://support.google.com/nexus/answer/2844832?hl=en). These steps were written for devices on Android 9+ but are very close for older supported devices.
```
TomBrien:
```suggestion
If you are using a self-signed certificate on Android then you may get stuck at a blank screen after entering and/or selecting your Home Assistant instance. In order to correct this issue you will need to make sure the URL is valid and that you import the certificate into Android's Trusted Certificates. Steps to perform this can be found [here](https://support.google.com/nexus/answer/2844832?hl=en). These steps were written for devices on Android 9+ but are very close for older supported devices.

```

dshokouhi:
@TomBrien are we ok with mentioning 0.106 in the step about sensor naming? I know we probably will want to reword it once the release is out on Wednesday but wanted to make sure the information was correct :)

TomBrien:
I’m fine with mentioning it. Beta is out now and I think it’s good info to publicise.

PR #192, opened 2020-02-24
Add note about disabling location and entity

dshokouhi:
For all users its important to mention that location needs to be enabled in order to get sensor updates.  We should also mention about disabling the entity in the registry if they wish to stop updates.


docs/core/location.md
```diff
@@ -19,7 +19,7 @@ Location updates are sent from your device to Home Assistant in a number of situ
...
+Depending on your set up, location data is sent directly from your phone to your Home Assistant instances or via the Home Assistant Cloud Service. This will depend on the URLs specified in the Connection section of the App Configuration menu. Location data is not sent via any other servers or organisations. Of course, if you decide not grant the Home Assistant Companion App location permission or if you subsequently remove the location permissions (![iOS](/assets/apple.svg) Settings>Privacy>Location Services or ![android](/assets/android.svg) Settings>Privacy>Permissions), no location data will be sent from your device to Home Assistant. Android users must also be aware that none of the other [sensors](sensors.md) will work if location is disabled. An alternative is to disable the `device_tracker.<device_name>` entity from the [entity registry](https://www.home-assistant.io/integrations/config/#entity-registry)
 **Check this is true for notification updates**
```
dshokouhi: Is this something we need to keep or was it meant to be a reminder for something else? :)

TomBrien: I see kill that line for now as it clearly shouldn’t be there. We should check but I suspect iOS has the same with no sensor data being sent if location is disabled. I can’t test sadly as iOS 13.4b2 is breaking sensor updates anyway

TomBrien:
```suggestion
```

dshokouhi: It looks like sensor updates on iOS are also tied to location.

https://companion.home-assistant.io/docs/core/sensors

```
It is important to know that these sensors are only updated when a location is pushed to Home Assistant or the web view is refreshed.
```

So I will go ahead and make my notes a bit more general now :)

PR #199, opened 2020-03-13
Added documentation on device_tracker attributes

Hypfer:
Since I was confused about what unit the `speed` value might be, I've updated the documentation according to the current code.

There might be iOS device things missing since I don't own any. Please check

I've used the code as a reference
https://github.com/home-assistant/home-assistant-iOS/blob/396f82c5f25b9a4d28790f8382aa3f09ef86aedd/Shared/API/Models/DeviceTrackerSee.swift#L126

### DIFF
--- a/docs/core/location.md
+++ b/docs/core/location.md
@@ -56,7 +56,7 @@ The newly created `device_tracker` entity may provide some of the following attr
 | speed             | meters/second |
 | vertical_accuracy | meters        |
 | floor             | floors  ![iOS](/assets/apple.svg)      |
-| timestamp         | Date          |
+| timestamp         | Date    ![iOS](/assets/apple.svg)      |
 
 If you want to know more about the specifics of these attributes, please refer to the relevant documentation of your operating system:
 
--- a/docs/core/location.md
+++ b/docs/core/location.md
@@ -55,7 +55,7 @@ The newly created `device_tracker` entity may provide some of the following attr
 | course            | degrees       |
 | speed             | meters/second |
 | vertical_accuracy | meters        |
-| floor             | floors        |
+| floor             | floors  ![iOS](/assets/apple.svg)      |
 | timestamp         | Date          |
 
 If you want to know more about the specifics of these attributes, please refer to the relevant documentation of your operating system:
--- a/docs/core/location.md
+++ b/docs/core/location.md
@@ -43,6 +43,27 @@ automation:
           entity_id: light.frontdoor
 ```
 
+## Entity attributes
+The newly created `device_tracker` entity may provide some of the following attributes depending on your operating system.
+
+| Name              | Unit          |
+|-------------------|---------------|
+| latitude          | degrees       |
+| longitude         | degrees       |
+| gps_accuracy      | meters        |
+| altitude          | meters        |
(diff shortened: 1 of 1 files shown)


docs/core/location.md
```diff
@@ -43,6 +43,27 @@ automation:
...
+| Name              | Unit          |
+|-------------------|---------------|
+| latitude          | degrees       |
+| longitude         | degrees       |
+| gps_accuracy      | meters        |
+| altitude          | meters        |
+| course            | degrees       |
+| speed             | meters/second |
+| vertical_accuracy | meters        |
+| floor             | floors        |
```
dshokouhi:
```suggestion
| floor             | floors  ![iOS](/assets/apple.svg)      |
```

Lets make sure to clarify which ones are iOS only, looks like only `floor` and `timestamp` are.  I think the rest are supported by android and iOS

docs/core/location.md
```diff
@@ -43,6 +43,27 @@ automation:
...
+| timestamp         | Date          |
```
dshokouhi:
```suggestion
| timestamp         | Date    ![iOS](/assets/apple.svg)      |
```

dshokouhi:
Hmm I am not sure why the checks are not passing now, seems they are stuck.

@TomBrien can you take a look at why netlify is stuck?  Also can you check that we have all the iOS attributes too? :)

TomBrien:
So build seems fine now. I'll check the rest now. Just need to find a charger...

docs/core/location.md
```diff
@@ -43,6 +43,27 @@ automation:
...
+|-------------------|---------------|
+| latitude          | degrees       |
+| longitude         | degrees       |
+| gps_accuracy      | meters        |
+| altitude          | meters        |
+| course            | degrees       |
+| speed             | meters/second |
+| vertical_accuracy | meters        |
+| floor             | floors  ![iOS](/assets/apple.svg)      |
+| timestamp         | Date    ![iOS](/assets/apple.svg)      |
```
TomBrien:
```suggestion
| `source`   |  _None_  |
| `battery_level`  |  percentage  |
| `latitude`          | degrees       |
| `longitude`         | degrees       |
| `gps_accuracy`      | meters        |
| `altitude`          | meters        |
| `course`            | degrees       |
| `speed`             | meters per second |
| `vertical_accuracy` | meters        |
| `floor`             | floors  ![iOS](/assets/apple.svg)      |
```

docs/core/location.md
```diff
@@ -43,6 +43,27 @@ automation:
...
+
+If you want to know more about the specifics of these attributes, please refer to the relevant documentation of your operating system:
+
+[Android](https://developer.android.com/reference/android/location/Location) or
+[iOS](https://developer.apple.com/documentation/corelocation/cllocation)
+
+
 ## Location tracking when outside a Home Assistant zone
```
TomBrien:
```suggestion

## Location tracking when outside a Home Assistant zone
```

docs/core/location.md
TomBrien: Should also add a note that `course` and `speed` are not reported when not moving and thus would be `unavailable`

PR #218, opened 2020-05-05
Clarify how android works with critical notifications

dshokouhi:
There was some confusion about whats supported on Android here.  Many users are looking to see if the phone will ring even when placed on silent.  Unfortunately this is not supported but users can override Do Not Disturb with notification channels so I put a note about it.


TomBrien:
@dshokouhi should we merge this?

dshokouhi:
@TomBrien yea I forgot to merge it earlier :)

PR #219, opened 2020-05-15
Add _redirects

TomBrien:
Add a redirect file to catch 404'd links in the iOS app


TomBrien:
Fixes home-assistant/iOS#588

Issue #176, opened 2020-02-14
Harmonizing and improving notifications across platforms


robbiet480:
_Please note, I wrote this while very tired and lacking sleep. It will receive updates over the next few days._

Right now, we live in a similar yet divided world. Home Assistant Companion (hereinafter HAC) for iOS came first and had many years to mature before HAC Android was even a glint in our eyes. Due to that time and somewhat limited thought for the future, iOS made some decisions that don't work well in the Android world. Android is starting to make decisions that will also affect iOS. We should try to cut down on as many differences between the two apps as possible. This will allow us to simplify our documentation but more importantly make it easier for users in households with both Androids and iPhones, in addition to making config sharing easier.

## Notifications

The first place to start with is notifications. Right now, both apps support different functionalities. Some functionalities exist because of work developers have done, some will never exist due to OS specific limitations. As much as possible though, we should make it possible for a user to send a notification and 90% of the time it will appear similar across platform. The way that we do this is deriving a schema that both apps can hew to while also providing allowances for platform specific functionality that users want.

As a first step towards that goal, I have come up with this:

```json
{
	"push_token": "",
	"registration_info": {
		"app_id": "io.robbie.HomeAssistant.beta",
		"app_version": "2.0.0 (68)",
		"os_version": "13.1.3"
	},
    "id": "uuid",
    "actions": [
        {
            "title": "Action 1",
            "id": "action_one"
        }
    ],
	"notification": {
        "message": "The front door is opened",
        "title": "Front Door",
		"badge": 5,
		"image": "https://github.com/home-assistant/home-assistant-assets/blob/master/logo-round-192x192.png?raw=true",
		"tag": "",
		"click_action": "",
		"android": {
			"priority": "",
			"ttl": "",
			"sticky": true,
			"visibility": "VISIBILITY_UNSPECIFIED",
			"local_only": true,
			"icon": "ics_launcher",
			"color": "#000000",
			"channel_id": "default",
			"ticker": "Hello World",
			"event_time": "2020-02-14T07:30:21Z",
			"notification_priority": "PRIORITY_DEFAULT",
			"default_sound": "",
			"default_vibrate_timings": true,
			"default_light_settings": true,
			"vibrate_timings": ["1s", "3.5s"],
			"light_settings": ""
		},
		"ios": {
			"image": {
                "content_type": "image/png",
                "hide_thumbnail": true
			},
			"subtitle": "TwelveTwelve",
			"sound": {
				"name": "default",
[...]

PR #243, opened 2020-06-10
Lots of new android notification improvements

dshokouhi:
We have several new properties now supported in the Android app since https://github.com/home-assistant/android/pull/612 was merged 🎉

- Icon
- LED Color
- Grouping
- Channel importance
- Vibration Pattern
- Subject
- Timeout
- HTML formatting for `message`
- Persistent notifications

I don't think any of these are supported by iOS at this time but will let the iOS guys comment on that :)

CC: @hasenbolle, I tried to capture everything you had written in your PR and even borrowed parts of it :) Let me know if I captured everything. Since the changes are now live in the beta channel we like to update the docs early to help out those users.

### DIFF
--- a/docs/core/index.md
+++ b/docs/core/index.md
@@ -162,14 +162,44 @@ Not all features are supported by Android at the moment but eventually most feat
       <td>✅</td>
     </tr>
     <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-grouping">Grouping</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-message-html-formatting">HTML Formatting</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-icon">Icon</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
       <td><a href="/docs/notifications/notification-attachments">Image</a></td>
       <td>✅</td>
       <td>✅</td>
     </tr>
     <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-channel-importance">Importance</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-led-color">LED Color</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
       <td><a href="/docs/notifications/notifications-basic">Message</a></td>
       <td>✅</td>
       <td>✅</td>
+    </tr>
+    <tr>
+      <td><a href="/docs/notifications/notifications-basic#persistent-notification">Persistent</a></td>
+      <td>✅</td>
+      <td></td>
     </tr>
     <tr>
       <td><a href="/docs/notifications/notifications-basic#controlling-how-a-notification-is-displayed-when-in-the-foreground">Presentation</a></td>
@@ -197,6 +227,11 @@ Not all features are supported by Android at the moment but eventually most feat
       <td></td>
     </tr>
     <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-subject">Subject</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
       <td><a href="/docs/notifications/notifications-basic#subtitle">Subtitle</a></td>
       <td></td>
       <td>✅</td>
@@ -207,9 +242,19 @@ Not all features are supported by Android at the moment but eventually most feat
       <td>✅</td>
     </tr>
     <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-timeout">Timeout</a></td>
+      <td>✅</td>
+      <td></td>
+    </tr>
+    <tr>
       <td><a href="/docs/notifications/notifications-basic">Title</a></td>
       <td>✅</td>
       <td>✅</td>
+    </tr>
+    <tr>
+      <td><a href="/docs/notifications/notifications-basic#notification-vibration-pattern">Vibration Pattern</a></td>
+      <td>✅</td>
+      <td></td>
     </tr>
     <tr>
       <td><a href="/docs/notifications/notification-attachments">Video</a></td>
--- a/docs/core/index.md
+++ b/docs/core/index.md
@@ -162,11 +162,6 @@ Not all features are supported by Android at the moment but eventually most feat
       <td>✅</td>
     </tr>
     <tr>
-      <td><a href="/docs/notifications/notifications-basic#notification-grouping">Grouping</a></td>
-      <td>✅</td>
-      <td></td>
-    </tr>
-    <tr>
       <td><a href="/docs/notifications/notifications-basic#notification-message-html-formatting">HTML Formatting</a></td>
       <td>✅</td>
       <td></td>
@@ -238,7 +233,7 @@ Not all features are supported by Android at the moment but eventually most feat
     </tr>
     <tr>
       <td><a href="/docs/notifications/notifications-basic#thread-id-grouping-notifications">Threads</a></td>
-      <td></td>
+      <td>✅</td>
       <td>✅</td>
     </tr>
     <tr>
--- a/docs/notifications/attachments.md
+++ b/docs/notifications/attachments.md
@@ -33,7 +33,7 @@ automation:
             hide-thumbnail: false
 ```
 
-Notes:
+Notes ![iOS](/assets/apple.svg):
 *   The thumbnail of the notification will be the media at the `url`.
 *   The notification content is the media at the `url`.
 *   Attachment can be used with custom push notification categories.
@@ -51,7 +51,10 @@ automation:
         message: "Something happened at home!"
         data:
           image: "https://github.com/home-assistant/home-assistant-assets/blob/master/logo-round-192x192.png?raw=true"
-```           
+```
+
+Notes ![android](/assets/android.svg):
+*   If you are setting the [`icon_url`](basic.md#notification-icon) and `image` property then only the image will be displayed on the device.
 
 
 ## Example
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -318,7 +318,7 @@ automation:
 ### Notification Channel Importance
 
 ![android](/assets/android.svg)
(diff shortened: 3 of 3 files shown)


dshokouhi:
So based on the official android docs and my testing I added a note about how channels work. In my testing with the vibration pattern I was able to change it once set but the default still applied if not given.  I don't have a device with a LED so I can't test my note there.

```
If you'd like to further customize your channel's default notification behaviors, you can call methods such as enableLights(), setLightColor(), and setVibrationPattern() on the NotificationChannel. But remember that once you create the channel, you cannot change these settings and the user has final control of whether these behaviors are active.
```
https://developer.android.com/training/notify-user/channels#CreateChannel

hasenbolle:
> So based on the official android docs and my testing I added a note about how channels work. In my testing with the vibration pattern I was able to change it once set but the default still applied if not given. I don't have a device with a LED so I can't test my note there.
> 
> ```
> If you'd like to further customize your channel's default notification behaviors, you can call methods such as enableLights(), setLightColor(), and setVibrationPattern() on the NotificationChannel. But remember that once you create the channel, you cannot change these settings and the user has final control of whether these behaviors are active.
> ```
> 
> https://developer.android.com/training/notify-user/channels#CreateChannel

Yes channel settings are only applied once on creation. After that users need to customize the channel in their device. The channel will be on each message overwritten, but the settings (vibration pattern, color leds etc) he used in the message  will be ignored, because the use might have changed the settings. The only thing that will work, is to lower the importance of a channel if the user hasn't changed the importance manually before.

hasenbolle:
Awseome work and thanks for writing the documentation. This will help the users a lot to deal with the notification system 👍 

Just some suggestions:
>If your device is on Android 8.0+ then some of the notification features may not work unless you specify a `channel`. The following properties will become the default for the `channel` if not provided: [`vibrationPattern`](#notification-vibration-pattern), [`ledColor`](#notification-led-color) and [`importance`](#notification-channel-importance).

All notifications features will work, if no channel is set. If the user do not set a channel, the General channel will be used. If this channel doesn't exist it will be created. So a channel is always used, regardless of whether the user set a channel.
Default values if not set are:
- Importance: `Default` which means `Default notification importance: shows everywhere, makes noise, but does not visually intrude.`
- Vibrtaion Pattern:  Vibration disabled
- LED Color: LED disabled
Tested this on a OnePlus 5T. Maybe some other manufactures have other defaults.

> Depending on when you installed the app you may want to send `remove_channel` to `channel: Default Channel` to clean up the old default channel:

The user should send `remove_channel` to channel `default`.

> Notification Channel Importance
The `channel` property is required for this feature to work.

The channel property is not required to work with the properties `importance`, `ledColor`, `vibrationPattern`  because the default channel is always used (The `General` channel).

> You can set the icon for a notification by providing the `icon_url`. The URL must be publicly accessible much like the `image` property. It is important to note that if you set the `image` then Android will not show the icon for the notification.

The `icon_url` can also point to a relative path on the system. 
`/local/community/bilder/post.png`
Like the `image` property also does.

Also if you set the `image` property in addition to the `icon_url` property. Then the `image` property will overwrite the icon in the message. So the message will be shown with the big image and with the image as the icon.

dshokouhi:
> All notifications features will work, if no channel is set. If the user do not set a channel, the General channel will be used. If this channel doesn't exist it will be created. So a channel is always used, regardless of whether the user set a channel.
> Default values if not set are:
> 
> * Importance: `Default` which means `Default notification importance: shows everywhere, makes noise, but does not visually intrude.`
> * Vibrtaion Pattern:  Vibration disabled
> * LED Color: LED disabled
>   Tested this on a OnePlus 5T. Maybe some other manufactures have other defaults.

So if the user already used the General channel and left the vibration pattern and LED color disabled then won't they need to create a new channel for a vibration pattern to take effect? I think for most users the general channel will exist before they start playing with this feature.  When I did my testing I didnt see any vibration until I created a new channel.

> The channel property is not required to work with the properties `importance`, `ledColor`, `vibrationPattern` because the default channel is always used (The `General` channel).


So we just need to inform the user that they can only lower the importance correct?  This whole set once thing is a little confusing around some of these options :) I think most users may hit that issue during testing due to the default options that get set the first time the channel is created.



> The `icon_url` can also point to a relative path on the system.
> `/local/community/bilder/post.png`
> Like the `image` property also does.
> 
> Also if you set the `image` property in addition to the `icon_url` property. Then the `image` property will overwrite the icon in the message. So the message will be shown with the big image and with the image as the icon.

I will get this added in my next update

hasenbolle:
> So if the user already used the General channel and left the vibration pattern and LED color disabled then won't they need to create a new channel for a vibration pattern to take effect? I think for most users the general channel will exist before they start playing with this feature. When I did my testing I didnt see any vibration until I created a new channel.

That is correct. In this case, they need to create a new channel with the notification message or change the setting of the `general` channel in the android channel settings, to take effect of `ledColor`,`importance `or `vibrationPattern `settings.

For me i had a similar experience with the channels, also before my pull req. I've send a notification to my phone, and didn't get  a popup nor vibration, because the default values are no popup and no vibration. I had to go to the channel settings, and set those settings manually. Maybe the default values of a newly created channel should then be importance= high and vibration = on. Which means: Show a popup, sound and enable vibration.

> So we just need to inform the user that they can only lower the importance correct? This whole set once thing is a little confusing around some of these options :) I think most users may hit that issue during testing due to the default options that get set the first time the channel is created.

They can only lower importance if they didn't changed the importance manually before. Yes it's kind of hard to understand why the for eg. the ledColors didn't changed even though they set it in the message.

Nevertheless i think, google did this the right way how they handle channels in android. It would be crazy if the app can overwrite the settings which the user set before. 

Maybe the properties should be renamed to `channelImportance`, `channelLedColors`, `channelVibrationPattern `to indicate that those settings are used for the channel. Now we have to tell the user that those properties are only used on channel creation. But the problem here is, that the importance property is also used for the priority setting for pre Oreo devices. On pre Oreo this setting can be changed every time, because in pre Oreo there are no channels :/

chriss158:
Just to avoid confusion. Renamed my account from hasenbolle -> chriss158 ^^

chriss158:
Did some tests. 
It would be possible to change the settings for importance, ledColor and vibrationPattern with every message. But it is kind of a brutal way. I delete the existing channel before every message (And therefore every custom setting which the user taken in the general channel). Then the channel will be recreated with the new settings. Maybe this would be a solution for the general channel and for all other user created channels the current behavior is used.

Btw: should we switch back to the pull req. of the app for such kind of discussion? Because this pull req is about the documentation.

dshokouhi:
> Did some tests.
> It would be possible to change the settings for importance, ledColor and vibrationPattern with every message. But it is kind of a brutal way. I delete the existing channel before every message (And therefore every custom setting which the user taken in the general channel). Then the channel will be recreated with the new settings. Maybe this would be a solution for the general channel and for all other user created channels the current behavior is used.
> 
> Maybe it is also possible to copy the old values of the channel to the new channel. This would solve all our problems, right? I will check this.
> 
> Btw: should we switch back to the pull req. of the app for such kind of discussion? Because this pull req is about the documentation.

Yea we should discuss that in the android repo, maybe under an enhancement since things are already merged.  I don't think recreating the `general` channel makes sense because then if the user customized it with a ringtone or whatnot we would lose that customization.  I think as long as we tell users what to expect they will be ok with what we decide. Giving them more control is always better :)

chriss158:
> Maybe it is also possible to copy the old values of the channel to the new channel. This would solve all our problems, right? I will check this.

This didn't work because some fields like bypass DND are not changeable by apps. So this was 
not an option anyway.

> Yea we should discuss that in the android repo, maybe under an enhancement since things are already merged. I don't think recreating the general channel makes sense because then if the user customized it with a ringtone or whatnot we would lose that customization. I think as long as we tell users what to expect they will be ok with what we decide. Giving them more control is always better :)

Ok then i think it's the best to stick with the standard behavior of android and leave everything as it is. As long the documentation gives a hint about this behavior it should be good. 👍 

One thing about the documentation:
> When you are setting the `channel` for your notification you also have the option to set the `importance` for the `channel` per notification. Possible values for this property are `high`, `low`, `max`, `min` and `default`. To learn more about what each value does see the [FCM docs](https://developer.android.com/training/notify-user/channels#importance). This feature is only for devices on Android 8.0+.

The `importance `property is used on pre oreo devices as priority. Which should work exactly the same way as the importance property for the channels in Oreo and higher.  So the importance feature/property can be used in pre Oreo devices
Only the channel property is useless for a pre oreo user.

TomBrien:
Is [notifications/basic](https://github.com/home-assistant/companion.home-assistant/blob/721e1524fe6525c39857166fcba8c7c0c46ae7d1/docs/notifications/basic.md) getting a bit too large to digest now? Maybe when this is merged we should look at splitting it out

dshokouhi:
@TomBrien I was just thinking about that myself. Starting to get a lot of additions.

TomBrien:
Given that at the moment the Android and iOS notifications work rather differently I wonder if we should split them out. I think the overlap is limited to pretty much the most basic example sadly?

dshokouhi:
I think technically `message` `topic` and `notify.group` would work for now since they work exactly the same.  The rest of it each platform has its own tidbits.  Maybe as https://github.com/home-assistant/companion.home-assistant/issues/176 starts to get implemented we can look into adding them to the same page. We also need to consider the other subpages too like for critical alerts and attachments.  Probably a good time to talk about the structure as both platforms mature more too.

dshokouhi:
Going to merge this now to get the info out there

joax01:
It would be great if someone could explain a how pictures (in portrait mode) like camera snapshots yaml coded into the notification.  A problem when expanding a notification image to full size, only part of the picture is shown. For example a security cam create portrait picture and the “full size” image is expanding by width, not height. There is a need of a way to control the full portrait mode picture in the notification.  As a general solution to this problem scaling support to Notification Icon/Images and Action bar Icon/images sizes when creating files would improve to control of image resolution in the complete "production chain" starting in the esphome camera and finally processed in home assistant services. In Android, in general, there is a need to handle 5 image versions to cover all the platforms available. The table below shows the pixel densities and icon sizes for the different types of user interface icons used by Android. How is this supported?:
 	Scale           	1 x	1.5 x	2 x	3 x	4 x
DPI	               ~ 160 dpi	~ 240 dpi	~ 320 dpi	~ 480 dpi	~ 640 dpi
App Launcher Icons	48 px	72 px	96 px	144 px	192 px
Action bar Icons	32 px (24px inset)	48 px	64 px	96 px	128 px
Small / Contextual Icons	16 px (12px inset)	24 px	32 px	48 px	64 px
Notification Icons	24 px (22px inset)	36 px	48 px	72 px	96 px

PR #244, opened 2020-06-11
Add notification links

TomBrien:
You can include a link in the message field so let's tell people about that!

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -23,6 +23,7 @@ The mobile_app platform provides many enhancements to the simple notification ge
 
 ### Including Links
 
+![iOS](/assets/apple.svg)
 If you include a URL in the `message:` of your notification, the link will be clickable when the notification is delivered. For example::
 
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/notifications/basic.md
```diff
@@ -21,6 +21,25 @@ The mobile_app platform provides many enhancements to the simple notification ge
 
 ## Enhancing basic notifications
 
+### Including Links
+
+If you include a URL in the `message:` of your notification, the link will be clickable when the notification is delivered. For example::
```
dshokouhi, line 26: Just tested this against the latest beta version of the android app.  The link only renders when I use the proper HTML tag `a` however it does not take the user to the link, it takes them to the HA app.  I think for android they will need to use either actionable notifications or click action as those work as expected.

TomBrien: Cheers will flag this as iOS only for now

docs/notifications/basic.md
```diff
@@ -21,6 +21,26 @@ The mobile_app platform provides many enhancements to the simple notification ge
...
+      ...
+    action:
+      service: notify.mobile_app_<your_device_id_here>
+      data:
+        message: "Touchdown 🚀: https://www.youtube.com/watch?v=l5I8jaMsHYk"
+```
+
+If your device has has an app associated with the domain, the link swill open in that app, otherwise the default browser is used. To make the link clickable you may had to tap and hold the notification.
+
+![android](/assets/android.svg) Android users can achieve this through the use of [actionable notifications](/docs/notifications/actionable-notifications#building-automations-for-notification-actions).
```
dshokouhi, line 42:
```suggestion
![android](/assets/android.svg) Android users can achieve this through the use of [actionable notifications](/docs/notifications/actionable-notifications#building-automations-for-notification-actions) or [click action](#notification-click-action).
```

PR #248, opened 2020-06-13
Add iOS app link redirects

TomBrien:
At the moment the app UI has several hard-coded links which can fall out of data easily requiring a new app release. This PR provides a resource to replace the hard coded links.

I've included all the UI links I could spot (quickly) in the source that are for the UI. I have not included auth or API endpoints. And I didn't include the base URLs for Facebook or Twitter because if they change we can presume the service is gone and the links will need removing anyway.

I have updated the companion doc links where needed but have left the What's New link for now as I don't think that is actually done anywhere. I suppose the docs site blog would be the best place?


robbiet480:
- Whats the deal with language here? I'd rather build logic into the apps to send users to the proper language docs vs having keys for each language.
- What URL will this be available at once merged?

TomBrien:
I thought you’d asked for language in the keys. Probably my mistake, wasn’t sure if you wanted to future proof. 

It’d be available at https://companion.home-assistant.io/app-links-ios.json

dshokouhi:
This is great! I wonder if the android app can benefit from this, currently we don't link to the docs from within the app.

TomBrien:
This is just a proposal at the moment. We might actually just use the site to host generic URLs and the redirect those with the `_redirects` file. Either way I'm sure we can use the same solution for Android.

homeassistant:
Hi @dependabot-preview[bot], 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#248).

Thanks!

PR #252, opened 2020-06-16
Update theming to reflect changes in iOS

TomBrien:
We now no longer need to use a service call to change the status bar (and other elements). Also have a slight change to the variables used. 

Quick check over from someone on the Android side would be great.


docs/integrations/theming.md
```diff
@@ -3,10 +3,20 @@ title: Theming
 id: 'theming'
 ---
 
-The app attempts to match the theme that the frontend is currently using. Theme keys that the app will use are:
+The process needed to get the app to match the Home Assistant theme is different depending on the app version used.
+
+## Android app and iOS apps after version 2020.2
```
dshokouhi, line 8: for android we still need to use `frontend.set_theme` that hasn't changed for us yet I dont think

docs/integrations/theming.md
```diff
@@ -3,10 +3,20 @@ title: Theming
...
 - `primary-background-color` for the background color of the web view ![iOS](/assets/apple.svg)
-- `primary-color` for the status bar background color ![iOS](/assets/apple.svg) ![android](/assets/android.svg)
-- `text-primary-color` for the pull to refresh tint color ![iOS](/assets/apple.svg)
+- `app-header-background-color` for the status bar background color ![iOS](/assets/apple.svg)
+- `primary-color` for the status bar background color ![android](/assets/android.svg); for the pull-to-refresh control/spinner ![iOS](/assets/apple.svg)
+
+Colors should be specified in hex format (e.g. `#0099ff`) and defining element colors through variable names is not supported.
 
-Themes should update in real time. Only system themes, not user themes, are supported. Colors should be specified in hex format (e.g. `#0099ff`) and defining element colors through variable names is not supported.
+![android](/assets/android.svg) Only system themes, not user themes, are supported.
```
dshokouhi, line 22: I think this may actually not be true as the user can add any theme to their configuration and reload it.

docs/troubleshooting/setup.md
```diff
@@ -58,7 +58,7 @@ This is probably not an issue with the Companion App but more likely with Home A
...
+If you are using the iOS app prior to version 2020.2, to change the color of the status bar to match your Home Assistant theme, please use the [`frontend.set_theme`](https://www.home-assistant.io/components/frontend/#theme-automation) service instead of the dropdown menu in the Home Assistant profile page. Using the service will generate an event allowing the Companion App to detect the theme change and apply the correct color to the status bar. See the [theming](../integrations/theming.md) documentation for details of which keys are used. Note that colors must be specified as hex values (e.g. `#0099ff`) in your theme and specifying element colors through variable names is not supported.
```
dshokouhi, line 61: We should probably also mention that the android app needs the service call too

PR #277, opened 2020-07-07
Minor update to iTunes import instructions

SeanPM5:
Mostly wanted to fix one thing (step 9 in the iTunes instructions) but also did a couple other extremely minor tweaks in the process.


TomBrien:
Looks good to me

PR #281, opened 2020-07-13
iOS Notification URL update

SeanPM5:
Updated the docs with a nicer example of iOS URL's (as per Discord chat). Also pointed out some configuration options on how to change behavior of links.

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -36,8 +36,8 @@ automation:
       data:
         title: "Motion Detected in Backyard"
         message: "Someone might be in the backyard."
-      data:
-        url: /lovelace/cameras
+        data:
+          url: /lovelace/cameras
 ```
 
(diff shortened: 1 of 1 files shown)


docs/notifications/basic.md
```diff
@@ -34,10 +34,16 @@ automation:
     action:
       service: notify.mobile_app_<your_device_id_here>
       data:
-        message: "Touchdown 🚀: https://www.youtube.com/watch?v=l5I8jaMsHYk"
+        title: "Motion Detected in Backyard"
+        message: "Someone might be in the backyard."
+      data:
+        url: /lovelace/cameras
```
TomBrien, line 40:
```suggestion
        title: "Motion Detected in Backyard"
        message: "Someone might be in the backyard."
        data:
          url: /lovelace/cameras
```

I only have one view set up at the moment so cant check but I assume it should be this?

TomBrien:
Looks good!

PR #287, opened 2020-07-28
Update instructions for defining notification categories and actions for iOS in configuration.yaml

TomBrien:
Currently the iOS integration only exists to support people defining their notification categories in YAML, that's it nothing else! I am putting in a PR to create the API view used by the app to get this info from the mobile_app: instead of ios:. This PR documents the changes made in that.


TomBrien:
Parent is now home-assistant/ios#981

PR #295, opened 2020-08-03
Add info about android entity state widget

dshokouhi:
A new widget was just added to the beta builds so here is some info on it, along with some minor clean up since we now have 2 widgets.

https://github.com/home-assistant/android/pull/687


dshokouhi:
@colincachia FYI :) we normally try to update the companion site as features get merged so beta users can see whats new.  I think I captured everything including the important bit about 30 min updates for the new widget.

PR #296, opened 2020-08-04
Correct tasker steps

dshokouhi:
It was discovered in https://github.com/home-assistant/android/issues/680 that we may need to update our steps.  Will wait to merge this change after we get confirmation from the user that opened the bug.


dshokouhi:
Got confirmation from the user that the steps are now working as desired.

PR #299, opened 2020-08-06
Initial commit of android flavors

dshokouhi:
The android app is now offered as both a `full` or `minimal` version so this will highlight the differences between them.  While we are at it lets also mention how users can help test out unreleased features/fixes with the `debug` builds.


dshokouhi:
CC: @jbassett I think I captured the differences correctly right?

dshokouhi:
Should we add a new column to the comparison table for the `minimal` flavor to really show what they can/can't do?

JBassett:
You captured the differences.  Adding a table might be nice.  Hopefully we can bring the two in line so they offer the same feature set eventually.

dshokouhi:
@JBassett So its only battery and wifi connection sensor the `minimal` flavor gets for now right? Geocoded is obviously a play services one lol.

JBassett:
@dshokouhi that's correct

dshokouhi:
ok cool, will update and add to the table :) thanks for confirming!

dshokouhi:
@TomBrien when you get a moment can you see if the table looks alright with the new column? Not sure if there is any room for improvement there :)

TomBrien:
Looks fine in terms of screen width to me. We should probably `vertical-align: top` the `th` cells is all.

dshokouhi:
@TomBrien better now? For some reason I don't see a difference on my end lol

TomBrien:
> @TomBrien better now? For some reason I don't see a difference on my end lol

For record needed to be done like <td style="vertical-align: top;>` but it's tidier to just stick a class on the table and add the CSS rules

PR #323, opened 2020-09-04
Fix OS and other images in dark mode

TomBrien:
Docusaurus >2.0.0-alpha62 appends a random string on the end of images specified in markdown. This breaks the hacky way I was using CSS selectors to invert some images (mainly the OS images). This PR switches over to defining these images via HTML so we can define a `class` attribute.


TomBrien:
So I kinda forgot this draft existed and fixed in #339

PR #326, opened 2020-09-09
Add hint regarding the possible use of the history stats integration

Hypfer:


dshokouhi (approved):
Great addition, thank you!

PR #342, opened 2020-10-02
Add info about call number sensor

dshokouhi:
A new sensor was merged today that retrieves the call number of the current call, here are more details about it.


frenck (approved):
Thanks, @dshokouhi 👍

PR #343, opened 2020-10-02
Mention which sensors require permissions

dshokouhi:
The app does a good job of prompting the user for permissions before we enable a sensor as some require it.  We are now going to list what permissions are required for sensors that need them.


frenck (approved):
Thanks, @dshokouhi 👍 

Happy Hacktoberfest 🎉

PR #345, opened 2020-10-03
Update android start fresh steps

dshokouhi:
The restart after renaming the device is no longer needed.  If a user has more than 1 entry (due to an old bug) they should remove all entries.


TomBrien (approved):
Looks good.

I should catch any iOS one of these. Would be cool to reload the integration when deleting an entry

dshokouhi:
@TomBrien yea I had to add in the step to remove multiple integrations since some users are still hitting that old bug.  I agree the less the user needs to do the better.

PR #346, opened 2020-10-05
Add info about android migration failure event

dshokouhi:
We found a couple odd situations in Sentry where the app database would fail to migrate crashing the app and not allowing the user to open the app.

We are now catching those exceptions which occur during the migration and informing the user of the crash by posting a notification and also including an event so the admin can create their own automations.

What we saw was that in some cases the sensors table would exist but no rows existed, this data is not expected to be empty. 

We also saw a case where a blob was stored in the database, we don't store blobs so it must've been corrupted by some other means or just bad API data.

### DIFF
--- a/docs/integrations/app-events.md
+++ b/docs/integrations/app-events.md
@@ -5,8 +5,10 @@ id: 'app-events'
 
 ## Overview
 
+
+To help with running automations, such as clearing app icon badges, or other tasks you may wish to trigger based on app usage the Home Assistant Companion Apps will fire different events onto the Home Assistant [event bus](https://www.home-assistant.io/docs/configuration/events/) during certain situations.
+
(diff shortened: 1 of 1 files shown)


dshokouhi:
Wonder if we should include the actionable notifications, sharing and tag events on this page too? Would be in a future PR if we decide to do that.

TomBrien (approved):
My only comment here is that it seems this is pretty serious and requires user to re set a lot if things. I wonder if an event will be missed. Perhaps a pop up in app or a notification is justified here.

TomBrien:
Ok ignore me. I didn't read the first comment properly. Notification is sent too.

dshokouhi:
@TomBrien yea we had a bit of discussion around how to notify the user and the admin lol

PR #354, opened 2020-10-07
Mention how users can bypass ringer mode for notifications on android

dshokouhi:
Changes for: https://github.com/home-assistant/android/pull/1019


BigDoom:
Just updated the app via Google Play to version 3.0.0-full, which should include this change, but when testing it's not working. I just see another channel called 'alarm-stream' has been added into notifications under the Home Assistant app. 

My automation in HA: 
action:
  -   service: notify.mobile_app_phone
       data:
         message: "Some message"
         title: "Some title"
           data:
             ttl: 0
             priority: high
             channel: alarm_stream

This just gives me a silent notification when the action is triggered.

What am I doing wrong here?

dshokouhi:
@BigDoom please dont comment on merged PR's always create an issue in the appropriate repo. In this case I suggest you check the alarm stream level on your phone.  If your alarm volume is low then it won't make a sound, its dependent on the volume level you have set there.

PR #357, opened 2020-10-09
Add info about last notification sensor

dshokouhi:
Info about a new recently merged sensor: https://github.com/home-assistant/android/pull/1018


dshokouhi:
@TomBrien are you ok if we merge this or should we look into redoing things with this change? I am honestly still not sure on a best approach :(

TomBrien (approved):

TomBrien:
> I am honestly still not sure on a best approach :(


Like neither but yeah go for it. Getting the info out there is more important than tidying up a big page

PR #358, opened 2020-10-11
Notification commands

dshokouhi:
This PR creates a new page for notifications to focus on the commands that we can send to devices.  I opted to let `request_location_update` to stay on its own page since I believe a lot of people link to it still.  I can definitely move it here if we think it should be kept in the same page.  I wanted to create a new page as these commands are really different than your standard notification.

This PR also contains the docs for recently merged: https://github.com/home-assistant/android/pull/1024


docs/notifications/commands.md
```diff
@@ -0,0 +1,78 @@
...
+      ...
+    action:
+      service: notify.mobile_app_<your_device_id_here>
+      data:
+        message: "command_dnd"
+        title: "priority_only"
+```
+
+## Request Location Updates
+
```
TomBrien, line 56:
```suggestion

![Android](/assets/android.svg) ![iOS](/assets/iOS.svg)
```

Let's flag clearly that this is platform independent

TomBrien (approved):
Looks good :)

PR #359, opened 2020-10-11
Add beta feature flag

TomBrien:
Adds a CSS rule for labelling a feature as beta. Implementation is to just add:

```html
<span class="beta">text</span>
```

Where `text` can of course be anything, if the span is left empty then "BETA" will be used.


dshokouhi (approved):
Thank you!!!

PR #360, opened 2020-10-11
Add info about android power menu integration

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1027

Also makes use of the newly added beta tag :) will submit another PR to mark the current beta features properly later on.

### DIFF
--- a/docs/integrations/android-power-menu.md
+++ b/docs/integrations/android-power-menu.md
@@ -1,0 +1,15 @@
+---
+title: "Android Power Menu"
+id: 'android-power-menu'
+---
+
+<span class="beta">BETA</span><br /><br />
+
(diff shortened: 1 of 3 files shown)


docs/integrations/android-power-menu.md
```diff
@@ -0,0 +1,15 @@
+---
+title: "Android Power Menu"
+id: 'android-power-menu'
+---
+
+<span class="beta">BETA</span><br /><br />
```
TomBrien, line 6: Do you mean to have two line breaks here?

dshokouhi: I did only because it looked too close with only one:
image

compare to 2 breaks:
image

I am ok with leaving it as one 😄

TomBrien: I see what you mean. Let's go with this. I need to tidy up the ow flags as a whole still so can sort that down the line.

TomBrien (approved):
👍

PR #367, opened 2020-10-15
Add media browser example

dshokouhi:
We mention `/local` but we should also mention `/media` now that we have the media browser

Fixes: https://github.com/home-assistant/android/issues/1059


dshokouhi:
If an iOS team member has a moment can you please check that this works as expected? I think it will but would like to double check if we have to call this out.

TomBrien:
> If an iOS team member has a moment can you please check that this works as expected? I think it will but would like to double check if we have to call this out.

Will try to today :)

TomBrien:
I can't get it to work but I also can't get 

>Another acceptable format for the above is `/local/path/to/file.jpg`

To work. Is this meaning I should be able to have `url: "/local/some/path/to/file.jpg"` in the payload? That Doesn't seem to work on iOS. Does Android include the file (rather than just link) in the payload? I believe in iOS that we just send the source path and iOS grabs the image. So that path wouldn't work in iOS and the fact that media is protected by HA auth for `/media` is probably stopping me getting the image.

dshokouhi:
@TomBrien yea so for android what we do is we check the URL to see if its absolute or a relative path.  If it is a relative path we actually add the auth headers.  I suppose iOS does not do that yet but I thought it did.  So I assume also the Camera Proxy API does not work on iOS either?

zacwest:
Relative URLs for images should work and should provide authentication around the retrieval. I'll see if I can replicate this.

docs/notifications/attachments.md
```diff
@@ -7,7 +7,7 @@ Notifications may contain an image, video, or audio file attachment that is disp
...
 ![](/assets/android.svg) Android users can also use `/api/camera_proxy/camera.name` where `camera.name` is replaced by the entity ID of the camera you wish to use.
```
TomBrien, line 11:
```suggestion

![Android](/assets/android.svg) Android users can also use `/api/camera_proxy/camera.name` where `camera.name` is replaced by the entity ID of the camera you wish to use.
```

docs/notifications/attachments.md
```diff
@@ -7,7 +7,7 @@ Notifications may contain an image, video, or audio file attachment that is disp
...
 ![](/assets/apple.svg) iOS users can use the [camera stream in dynamic attachments](dynamic-content.md#camera-stream).
```
TomBrien, line 12:
```suggestion

![iOS](/assets/apple.svg) iOS users can use the [camera stream in dynamic attachments](dynamic-content.md#camera-stream).
```

docs/notifications/attachments.md
```diff
@@ -7,7 +7,7 @@ Notifications may contain an image, video, or audio file attachment that is disp
...
...hat the attachments are required to be accessible from the internet. If you plan to use [`camera.snapshot`](https://www.home-assistant.io/integrations/camera#service-snapshot), you will want to store the image in the `www` folder located in your Home Assistant [config directory](https://www.home-assistant.io/docs/configuration/). This will then expose the image to the internet so that you may use it in your notifications and receive them anywhere. The proper format for the URL is: `https://MyCustomHADomain/local/path/to/file.jpg`. Make sure to replace `MyCustomHADomain` with how you would access Home Assistant remotely in a browse, you will either use your [Nabu Casa remote URL](https://www.nabucasa.com/config/remote/) or the [base URL](https://www.home-assistant.io/integrations/http#base_url). Also, replace `path/to/file.jpg` with the actual file name and location you used in your [`camera.snapshot`](https://www.home-assistant.io/integrations/camera#service-snapshot) service call. Another acceptable format for the above is `/local/path/to/file.jpg` or if using the [`media_source` integration](https://www.home-assistant.io/integrations/media_source) `/media/local/path/to/file.jpg`.
```
TomBrien, line 10:
```suggestion
It is important to note that the attachments are required to be accessible from the internet. If you plan to use [`camera.snapshot`](https://www.home-assistant.io/integrations/camera#service-snapshot), you will want to store the image in the `www` folder located in your Home Assistant [config directory](https://www.home-assistant.io/docs/configuration/). This will then expose the image to the internet so that you may use it in your notifications and receive them anywhere. The proper format for the URL is: `https://MyCustomHADomain/local/path/to/file.jpg`. Make sure to replace `MyCustomHADomain` with how you would access Home Assistant remotely in a browse, you will either use your [Nabu Casa remote URL](https://www.nabucasa.com/config/remote/) or the [base URL](https://www.home-assistant.io/integrations/http#base_url). Also, replace `path/to/file.jpg` with the actual file name and location you used in your [`camera.snapshot`](https://www.home-assistant.io/integrations/camera#service-snapshot) service call. Another acceptable format for the above is `/local/path/to/file.jpg`. 

If you are using the [`media_source` integration](https://www.home-assistant.io/integrations/media_source) you can alternatively use `/media/local/path/to/file.jpg`. Using `media_source` has the advantage that access requires authentication headers (which Home Assistant provides to the companion app). This means the content is not publicly available.
```

TomBrien (changes requested):
We currently hackily use the image alt text in the CSS so need to include. Let's promote media_source harder as it's pretty assume privacy wise.

Also, I presume the OS specific bits were always supposed to be on separate lines.

TomBrien:
Ok after literally years of telling people to make sure they use `default_config` turns out I didn't in production 🤦‍♂️! This works great with iOS

TomBrien (approved):
Looks good to me! Apologies again for the fail on my part 😳

dshokouhi:
@TomBrien no problem :)

PR #373, opened 2020-10-26
Fix typo in iOS event example

dwradcliffe:
I'm not sure if this changed at some point but for me on 0.116.4 the event data item was `action_data`.


TomBrien:
Good spot confirmed here and you're quite right. My gut is that the Android change affected iOS. Your branch doesn't seem to be editable, can you update [Line 215](https://github.com/dwradcliffe/companion.home-assistant/edit/patch-1/docs/notifications/actionable.md#L215) to match.

TomBrien:
Looks great. Cheers for fixing this 👍

PR #374, opened 2020-10-26
Updates for recent android beta changes

dshokouhi:
*  New power menu features
*  New widget
*  Removal of Call Number sensor per Google Play Policy change

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -51,7 +51,6 @@ All sensors update during a periodic 15-minute interval and they will also updat
 | [Audio Sensors](#audio-sensors) | None | Several different sensors around different types of audio detection from the device. |
 | [Battery Sensors](#battery-sensors) | None | Several different sensors around the state of the devices battery. |
 | `sensor.bluetooth_connection` | [See Below](#bluetooth-sensor) | The state of the sensor will reflect the total number of connected bluetooth devices. |
-| `sensor.call_number` | None | The number of the current incoming/outgoing call. |
 | `sensor.do_not_disturb` | None | The state of do not disturb on the device. |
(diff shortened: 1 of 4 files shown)


docs/core/sensors.md
```diff
@@ -144,10 +143,6 @@ This sensors state will be the total number of connected bluetooth devices. The
 There will also be a binary sensor for the `bluetooth_state` that will represent whether or not bluetooth is turned on for the device. This sensor will update anytime the state of bluetooth changes.
 
 
-## Call Number Sensor
-![Android](/assets/android.svg) &nbsp;<span class="beta">BETA</span><br />This sensors state will reflect the number of the current incoming/outgoing call, the state will be `none` when there is no active call. This sensor works best when paired with the [Phone State Sensor](#phone-state-sensor) as it will update at the same time and uses the [TelephonyManager API](https://developer.android.com/reference/android/telephony/TelephonyManager?hl=en). This sensor requires the [Read Call Log permission](https://developer.android.com/reference/android/Manifest.permission#READ_CALL_LOG).
```
TomBrien, line 148: Not surprised Google dropped the ability to get the call number, sounds like a previous nightmare. Is it worth rather than deleting this chunk replacing it with a note for anyone running an old app build?

dshokouhi: Well this was actually a beta feature that never made it into the Play Store Beta so only a very small handful of users may have had it but it was already removed.  It only existed for like 10-20 beta builds before we had to pull it out.

One thing to point out too is that our new Last Notification sensor can actually do the same thing this one can as phone calls come in via notification and we can get the contact information from there :)

TomBrien: Oh yeah. There's that clever beta flag. Yeah burn it.

TomBrien (approved):

PR #381, opened 2020-11-03
Mention usage of alarm_stream_max for TTS notifications

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1139

~~Lets make sure to merge after the parent PR has been merged~~

Parent PR merged :)

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -547,3 +547,18 @@ By default Text To Speech Notifications use the Music stream so they will bypass
         data:
           channel: alarm_stream
 ```
+
+If you find that your alarm stream volume might be too low you can use `channel: alarm_stream_max` which will temporarily set the alarm stream volume to the max level, speak the notification and then revert back to the original volume level.
+
+```yaml
+  - alias: Notify Alarm Triggered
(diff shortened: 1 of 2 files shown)


dshokouhi:
In a future PR we should probably setup a PR template in this repo to match the companion repos? I think that would also be required for the bot to do its thing?

docs/notifications/basic.md
```diff
@@ -547,3 +547,18 @@ By default Text To Speech Notifications use the Music stream so they will bypass
         data:
           channel: alarm_stream
 ```
+
+If you find that your alarm stream volume might be too low you can use `channel: alarm_stream_max` which will temporarily set the alarm stream volume to the max level, speak the notification and then revert back to the original volume level.
```
TomBrien, line 551:
```suggestion
If you find that your alarm stream volume is too low you can use `channel: alarm_stream_max` which will temporarily set the alarm stream volume to the max level, play the notification and then revert back to the original volume level.
```
I don't know the notification but I think this might read more clearly. I might be totally wrong

TomBrien (approved):
Looks great to me.

PR #400, opened 2020-11-21
Add another example for TTS notifications

Mincka:
Add another example for TTS notifications to explain how to prevent notifications to be spoken based on the Ringer mode.

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -563,7 +563,7 @@ If you find that your alarm stream volume is too low you can use `channel: alarm
           channel: alarm_stream_max
 ```
 
-If you do not want the TTS notification to be spoken when the Ringer mode is not `normal` (to mimic the behavior of classic notifications), you need to prevent the TTS notification to be sent in the first place. This can be done by adding a condition in your automation that checks the state of the [Ringer Mode](https://companion.home-assistant.io/docs/core/sensors#audio-sensors). In the same way, to prevent the TTS to speak when Do Not Disturb is enabled or when you are in a call, you can use the [Do Not Disturb](https://companion.home-assistant.io/docs/core/sensors/#do-not-disturb-sensor) sensor and the Audio Mode sensor that are also both provided by the companion app.
+You may not want the TTS notification to be spoken in certain situations (e.g. if the Ringer mode is not `normal` or DND is enabled). This can be done by adding a condition in your automation that checks the state of [other sensors](https://companion.home-assistant.io/docs/core/sensors). Few examples are presented below:
 
 ```yaml
   - alias: Notify of Motion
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @Mincka, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#400).

Thanks!

docs/notifications/basic.md
```diff
@@ -563,6 +563,22 @@ If you find that your alarm stream volume is too low you can use `channel: alarm
           channel: alarm_stream_max
 ```
 
+If you do not want the TTS notification to be spoken when the Ringer mode is not `normal` (to mimic the behavior of classic notifications), you need to prevent the TTS notification to be sent in the first place. This can be done by adding a condition in your automation that checks the state of another sensor, like the Ringer Mode (`audio_sensor`) or the Audio Mode (`audio_mode`).
```
dshokouhi, line 566: So we actually renamed `audio_sensor` to `ringer_mode` when we did the sensor breakup so to new users they will see `ringer_mode`. Also `audio_mode` is not the correct sensor for this as it does not relate to the actual ringer mode.

Lets also add a link to the sensor as well so they know its part of the app.  

https://companion.home-assistant.io/docs/core/sensors#audio-sensors

dshokouhi (changes requested):
Lets just make sure we follow what's outlined in the docs and clarify to the user which sensor they should use.

docs/notifications/basic.md
Mincka: Don't you think both sensors are relevant for this use case? I think it could be annoying to have the TTS speaks when the phone is ringing or even worse when you are in a call. That why using the `audio_mode` to check if the phone is "idle" (state is `normal`) makes sense to conditionally send the TTS command. That's just another idea, then people can check the documentation to understand the difference.

```suggestion
If you do not want the TTS notification to be spoken when the Ringer mode is not `normal` (to mimic the behavior of classic notifications), you need to prevent the TTS notification to be sent in the first place. This can be done by adding a condition in your automation that checks the state of the [Ringer Mode](https://companion.home-assistant.io/docs/core/sensors#audio-sensors) (`ringer_mode`). The [Audio Mode](https://companion.home-assistant.io/docs/core/sensors#audio-sensors) (`audio_mode`) can also be checked to prevent the TTS to speak when you are in a call.
```

dshokouhi: Ok makes sense :) lets just keep 1 link since it goes to the same place and drop the duplicate sensor names since it looks odd next to each other.  The audio sensors are indeed a good use here.

dshokouhi (approved):

PR #406, opened 2020-12-02
Mention android system webview for troubleshooting

dshokouhi:
Its a pretty important step to mention as it also solves issues users can see with the front end.  Also since we are linking to the actual app that they need to update might as well link to the HA app again in case the user wonders "which android app" :)


frenck (approved):

JBassett (approved):

TomBrien (approved):
👍

PR #407, opened 2020-12-08
Update basic.md

chpego:
the original seems to indicate that conditions are needed, there seems to be an oversight

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -569,7 +569,7 @@ You may not want the TTS notification to be spoken in certain situations (e.g. i
   - alias: Notify of Motion
     trigger:
       ...
-    action:
+    condition:
       - condition: state
         entity_id: sensor.<your_device_id_here>_ringer_mode # Only speak if the Ringer is normal (not vibrate or silent)
         state: normal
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @chpego, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#407).

Thanks!

dshokouhi (commented):
So the existing yaml is still valid, why do we need to change this?

TomBrien:
Current example (incorrectly) lists the conditions in the `action` block and not in a separate `condition` block

chpego:
> 
> 
> Current example (incorrectly) lists the conditions in the `action` block and not in a separate `condition` block

Thanks @TomBrien

dshokouhi:
Well you can still list conditions in the `action` section if you wish to start an `action` and then abort it later so it wasn't entirely incorrect there.  I have plenty of automations setup like that personally so the automation will still execute and respect the conditions.

TomBrien:
I guess the confusion could arise from the differences to <https://www.home-assistant.io/docs/automation/condition/>

dshokouhi (approved):

PR #415, opened 2020-12-20
Clarify map template example

### LINKED ISSUE #414: Clarification needed in dynamic notifcation docs
This should be a pretty quick fix.

On [this page](https://companion.home-assistant.io/docs/notifications/dynamic-content), it says:

`You may also use a device_tracker for the latitude and longitude coordinates like so: "{{states.device_tracker.<your_device_id_here>.attributes.latitude}}" but make sure to use data_template in that case.`

As the automation has two levels of data object, these docs should be clarified to specify which data object should be specified as a data_template: the first, the second, or both.

TomBrien:
We no longer need to use the `data_template` key so ditch that confusion (child key called data too). Currently due to typing of templates, requires a work around in app which is (will be) in next  beta

Fixes #414

### DIFF
--- a/docs/notifications/dynamic-content.md
+++ b/docs/notifications/dynamic-content.md
@@ -22,7 +22,13 @@ data:
 ```
 Be aware, that to send a map you must send a push `category` which has to be called `map`, `map1`, `map2`, `map3` or `map4`  otherwise you won't get the map delivered.
 
-You may also use a device_tracker for the latitude and longitude coordinates like so: `"{{states.device_tracker.<your_device_id_here>.attributes.latitude}}"` but make sure to use `data_template` in that case.
+You may also use a `device_tracker` for the latitude and longitude coordinates like so:
+
+```yaml
+    action_data:
+      latitude: "{{states.device_tracker.<your_device_id_here>.attributes.latitude}}"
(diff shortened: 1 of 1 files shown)


TomBrien:
I fully intend to submit a PR after this is merged removing this section as it's currently broken in core but can then revert that when core is fixed

docs/notifications/dynamic-content.md
```diff
@@ -22,7 +22,15 @@ data:
...
 
-You may also use a device_tracker for the latitude and longitude coordinates like so: `"{{states.device_tracker.<your_device_id_here>.attributes.latitude}}"` but make sure to use `data_template` in that case.
+### Providing coordinates using templates <span class="beta">BETA</span><br />
+
+You may also use a `device_tracker` for the latitude and longitude coordinates like so:
+
+```yaml
+    action_data:
+      latitude: "{{states.device_tracker.<your_device_id_here>.attributes.latitude}}"
+      longitude: "{{states.device_tracker.<your_device_id_here>.attributes.longitude}}"
```
frenck, line 32: This is not how we should access attributes, use `state_attr()` instead.

See: <https://www.home-assistant.io/docs/configuration/templating/#states>

frenck (changes requested):

frenck (approved):

PR #421, opened 2021-01-06
Update sensors documentation for macOS specifics

zacwest:


docs/core/sensors.md
```diff
@@ -49,7 +49,8 @@ On macOS, sensors update in the same situations as above as well as immediately
 | `binary_sensor.camera_in_use` | None | Whether a camera on the system is currently in use. |
 | `binary_sensor.microphone_in_use` | None | Whether a microphone on the system is currently in use. |
 | `sensor.displays` | `Display IDs`, `Display Names` | Requires 2021.2 or later. Number of displays connected to the device. |
-| `sensor.primary_display` | `Display ID` | Requires 2021.2 or later. Name of the current primary display, which is the display with the menu bar. |
+| `sensor.primary_display_id` | None | Requires 2021.2 or later. ID of the current primary display, which is the display with the menu bar. In the form of a UUID, for example `BE82E2E6-EA40-4963-93AD-A0BDC9D2F18F`. |
+| `sensor.primary_display_name` | None | Requires 2021.2 or later. Name of the current primary display, which is the display with the menu bar. |
```
TomBrien, line 53:
```suggestion
| `sensor.primary_display_id` | None | Requires Companion App version 2021.2 or later. ID of the current primary display, which is the display with the menu bar. In the form of a UUID, for example `BE82E2E6-EA40-4963-93AD-A0BDC9D2F18F`. |
| `sensor.primary_display_name` | None | Requires Companion App version 2021.2 or later. Name of the current primary display, which is the display with the menu bar. |
```

Now that Core uses calver too probably best to clarify exactly what we mean when giving a version number

TomBrien (changes requested):
Minor change to avoid potential confusion

docs/core/sensors.md
zacwest: went with the slimmer "app version" and got a few other places where this was used

TomBrien (approved):
Looks great and cheers for doing the others and saving me a PR to do them.

PR #429, opened 2021-01-13
Add info about android intent events

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1292

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -407,7 +407,7 @@ This sensors state will be the date and time of the last reboot from the device 
 ## Last Update Trigger Sensor
 ![Android](/assets/android.svg)
 
-For android this sensors state will reflect the [intent](https://developer.android.com/reference/android/content/Intent) of the most recent update sent.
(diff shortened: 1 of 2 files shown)


docs/core/sensors.md
```diff
@@ -407,7 +407,7 @@ This sensors state will be the date and time of the last reboot from the device
 ## Last Update Trigger Sensor
 ![Android](/assets/android.svg)
 
-For android this sensors state will reflect the [intent](https://developer.android.com/reference/android/content/Intent) of the most recent update sent.
+For android this sensors state will reflect the [intent](https://developer.android.com/reference/android/content/Intent) of the most recent update sent. Additionally the sensor offers settings to allow the user to receive [app events](../integrations/app-events.md) from other Android apps that broadcast an intent. Users can register for as many intents as they like, an event will be sent to Home Assistant once the intent has been received. Once you save an intent be sure to restart the application to register for the intent.
```
zacwest, line 410: seems like an unexpected place to put it

zacwest (approved):

docs/core/sensors.md
dshokouhi: Well the thought is that all registered intents we have come through to this sensor. In fact some times the state will even update to the intent the user entered. The reason for the event is that it can fire more than once and existing sensor logic is to not send a state update if it hasn't changed. I thought about making a new sensor but it's odd since it wouldn't be an actual sensor and if it were the state would be a duplicate of Last Update. If a user only wanted one intent it would never update.

PR #430, opened 2021-01-14
Remove duplicate battery attributes from level/state

zacwest:


TomBrien (approved):
Looks fine. Just waiting on home-assistant/iOS#1372

PR #434, opened 2021-01-17
Add notes about sensors with settings on iOS/macOS

zacwest:


TomBrien:
Builds fine for me locally. Probably a netlify thing. I'll try a gain in a little bit

TomBrien (approved):

PR #456, opened 2021-02-05
Mention android support for haptics

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1352

### DIFF
--- a/docs/integrations/haptics.md
+++ b/docs/integrations/haptics.md
@@ -3,18 +3,21 @@ title: Haptics
 id: 'haptics'
 ---
 
-![iOS](/assets/iOS.svg)<br />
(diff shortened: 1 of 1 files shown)


docs/integrations/haptics.md
```diff
@@ -3,18 +3,21 @@ title: Haptics
 id: 'haptics'
 ---
 
-![iOS](/assets/iOS.svg)<br />
```
zacwest, line 6: did you lose this ios header?

dshokouhi: I removed it since both apps will be supporting it now which is why I added the android beta note below the description.

zacwest: there's an iphone-specific description just below it, may pair well with an ios header similar to the android one below that

dshokouhi: ok I added it back....I don't know if it makes sense to keep both at the top since the android beta flag may confuse some iOS users if they are close to each other lol

zacwest (approved):

PR #459, opened 2021-02-09
Mention 3 finger swipe for quick bar

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1356

### DIFF
--- a/docs/integrations/android-webview.md
+++ b/docs/integrations/android-webview.md
@@ -1,7 +1,12 @@
 ---
-title: "Android WebView Links"
-id: 'android-webview-links'
+title: "Android WebView"
+id: 'android-webview'
 ---
+
+
+![Android](/assets/android.svg) &nbsp;<span class="beta">BETA</span><br />
+
+## Links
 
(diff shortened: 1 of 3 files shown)


docs/integrations/android-webview.md
```diff
@@ -20,3 +25,7 @@ This example will launch the barcode scanning app ready to scan via the Intent s
   name: Scan
   url: "intent://scan/#Intent;scheme=zxing;package=com.google.zxing.client.android;end"
 ```
+
+## Swipe Gestures
+
+The ![Android](/assets/android.svg) Android app has the ability to launch the [quick bar](https://www.home-assistant.io/docs/tools/quick-bar/) by detecting a 3 finger swipe down gesture. Initially the entity view will be shown to users and additional gestures will flip between the 2 quick bar views. The quick bar can be closed by simply pressing the back button. The quick bar can only be launched when inside the webview after you have logged in.
```
zacwest, line 31: "additional gestures" in the future or now?

zacwest (approved):

docs/integrations/android-webview.md
dshokouhi: Meaning that as long as the quick bar is open (and the text field does not have focus 😝 ) if the user keeps sending 3 finger swipe down gestures it will just flip between the 2 views.  All we are doing for now is sending the `e` key since an event doesn't exist lol

PR #460, opened 2021-02-09
Fix Activity intent examples

mdegat01:
Activity intent examples still showing the Intent Action string in the `group` field, now should be in the `tag` field


dshokouhi (approved):
Good catch!

PR #463, opened 2021-02-18
Fix an example with actionable notification url

denysdovhan:
I've spent too much time trying to figure out what I'm doing wrong here. Turns out it was a typo in an example. I hope this will save someone's time.

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -129,11 +129,12 @@ action:
   service: notify.mobile_app_<your_device_id_here>
   data:
     message: "Something happened at home!"
-    push:
-      category: "ALARM"
-    url: 
-      _: "/lovelace/cameras" # if the notification itself is tapped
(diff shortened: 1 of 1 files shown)


dshokouhi:
Do we need to worry about the second example just beneath as well? I don't use iOS so not sure about the expected format here.

TomBrien (changes requested):
@denysdovhan sorry for taking ages to get to this and thanks for fixing.

Can you update the next example https://github.com/home-assistant/companion.home-assistant/blob/3de4a5c6e24c2808356a9a232f5e099d6cb14a23/docs/notifications/actionable.md#L125-L136
to add the missing `data` key after `message`  

We should mention this won't work with actions which have `activationMode: background` (@zacwest can we workaround that if there is an URL defined for an action)

TomBrien (approved):
Looks great now. Thanks for sorting this 👍

denysdovhan:
Thank you!

PR #472, opened 2021-03-08
Add high accuracy zone docs

chriss158:
Docs for PR https://github.com/home-assistant/android/pull/1415


dshokouhi (approved):
Looks good to me, ok to merge when parent is merged.  If we make changes to the parent PR we may need to reflect them here.

PR #476, opened 2021-03-10
Update background explanation

TomBrien:
Updates the description of background fetch updates in iOS and explains why initial performance is poor and how to improve.

@zacwest can you just cast an eye over this to check there isn't anything massively untrue in it.


docs/troubleshooting/setup.md
```diff
@@ -142,7 +142,9 @@ If you have more than 1 device of the same model and you did not rename your dev
...
 
 ## Sensors are missing or not updating
-![iOS](/assets/iOS.svg) When the app is not in the foreground, sensor updates are tied to location updates so you need to make sure that location permissions are enabled for the device and app in order for sensor updates to be sent to the phone. Once you enable location for the app the sensors will show up and continue to update as long as the location is enabled on the device and app.
+![iOS](/assets/iOS.svg) When the app is not in the foreground, sensor updates are tied to location updates, so you need to make sure that location permissions are set to "Allow Always" in iOS settings.
+
+The app will also try so send updates in the background however the frequency of these is determined by iOS and is heavily throttled to protect battery life. iOS uses an internal metric, which is not visible to app developers, to prioritize background activities for apps. Apps which you use more will be allowed to do more in the background more frequently, this means that the more you use the Companion App, iOS will learn that the app is important to you and allow more frequent updates via background fetch.
```
frenck, line 147:
```suggestion
The app will also try to send updates in the background however the frequency of these is determined by iOS and is heavily throttled to protect battery life. iOS uses an internal metric, which is not visible to app developers, to prioritize background activities for apps. Apps which you use more will be allowed to do more in the background more frequently, this means that the more you use the Companion App, iOS will learn that the app is important to you and allow more frequent updates via background fetch.

```

zacwest, line 147: i wonder if it would be worth adding a "how to get charging state to update automatically" tutorial here around creating a personal automation in iOS 14, since that seems to be by-far the most requested flow

TomBrien: Good idea, will add that in a bit. Hmm what's the best way to share a shortcut? Screenshot?

TomBrien: Went textual as it's pretty minimal

zacwest (approved):

PR #480, opened 2021-03-24
Clarify requirement for intent and fix typo

aneisch:
Clarify that there is a requirement for Single Accurate Location sensor to be enabled for intents to work.

Fix typo in Android Location Settings section.


dshokouhi (approved):
Good catch!

aneisch:
> Good catch!

I watched logcat scroll by for far too long before realizing this :)

PR #481, opened 2021-03-25
Add info about battery temperature sensor

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/1453


docs/core/sensors.md
```diff
@@ -209,13 +209,14 @@ These sensors use the [AudioManager](https://developer.android.com/reference/kot
...
+The battery sensors listed below describe the state of the battery for a few different data points. The sensors icon reflects the charging status, and type of charging being used. The `battery_state`, `charger_type` and `is_charging` sensor will be updated when the device has a charger connected or disconnected. The `battery_health`, `battery_level` and `battery_temperature` sensor will be updated any time any of the other sensors get an update as well as when the device reports low battery and when it has recovered from the low battery alert. All of these sensors make use of [BatteryManager](https://developer.android.com/reference/android/os/BatteryManager?hl=en).
```
TomBrien, line 212:
```suggestion
The battery sensors listed below describe the state of the battery for a few different data points. The sensor's icon reflects the charging status, and type of charging being used. The `battery_state`, `charger_type` and `is_charging` sensor will be updated when the device has a charger connected or disconnected. The `battery_health`, `battery_level` and `battery_temperature` sensors will be updated any time any of the other sensors get an update as well as when the device reports low battery or when it has recovered from the low battery alert. All of these sensors make use of [BatteryManager](https://developer.android.com/reference/android/os/BatteryManager).
```

TomBrien (approved):
From my point of view it's good to go when the parent merges

PR #503, opened 2021-04-20
Fix anchors, update rate limit count, add clear_notification to iOS commands

zacwest:


docs/core/index.md
```diff
@@ -199,13 +199,13 @@ Not all features are supported by Android at the moment but eventually most feat
       <td><a href="/docs/notifications/notification-cleared">Cleared</a></td>
       <td>✅</td>
       <td></td>
-      <td></td>
+      <td>✅</td>
     </tr>
     <tr>
-      <td><a href="/docs/notifications/notifications-basic#notification-click-action">Click Action</a></td>
+      <td><a href="/docs/notifications/notifications-basic#opening-a-url">Opening a URL</a></td>
```
TomBrien, line 205: This should move down the list to keep the subsection alphabetised

TomBrien (changes requested):
One, pretty minor thing

TomBrien:
The feature list currently doesn't cover macOS at all. Should we add a column sometime for this. The iOS and macOS features are not identical?

TomBrien (approved):
Wow, I didn't mean the macOS column needed adding right now. Looks great!

PR #514, opened 2021-05-04
Update Siri Shortcuts docs on executing shortcuts

zacwest:


docs/integrations/siri-shortcuts.md
```diff
@@ -80,9 +72,42 @@ Shortcuts are deeply integrated into the OS. After creating one, you have numero
...
+        another_key: "another value"
+```
+
+When you tap on the notification to launch Home Assistant, it will redirect you to the Shortcuts app to perform the given Shortcut. Once you complete the Shortcut, it'll return you to Home Assistant and fire an event.
+
+:::note
+If the Shortcut doesn't require any input, it may appear as though the Shortcuts app wasn't launched at all. Check for the event being performed to see the results.
+:::
+
+The event fired is `ios.shortcut_run` with the result of the Shortcut with the following keys:
```
TomBrien, line 101: Is this only fired when the short cut is triggered from a notification?

zacwest: yeah

TomBrien (approved):

TomBrien:
Bonus points for the commit messages as it got later

PR #520, opened 2021-05-08
Update download badges

TomBrien:
Apple's link maker is dead and so our old download link was empty. Updated both badges from their respective store's current link makers and massaged a little to make them appear roughly equal in size


zacwest (approved):
I think the root page needs it too

zacwest:
Looks a bit off
image

TomBrien:
Yep literally spotted that as I re requested. Still had the local styling in Chrome dev tools

zacwest (approved):

PR #529, opened 2021-05-15
Add a notes for upgrading, watchOS and action count

zacwest:
- Upgrading to 2021.5 may require restarting the device for actions to work.
- watchOS requires the app installed for dynamic actions.
- There's not really a limit for actions anymore with dynamic actions, but the system UI stops allowing you to scroll around 10 or so, depending on the device size -- my 11 Pro can fit about 12 before I can't reach the later ones.


TomBrien (approved):
Looks good to me

PR #534, opened 2021-05-29
Add page anchors for critical notifications to link easily to different examples

dshokouhi:
Should make it easier to link to specific examples instead of saying things like "3rd example down"


TomBrien (approved):
Nice solution

PR #535, opened 2021-05-31
describe Android's high priority option

matejdro:
As discussed in https://github.com/home-assistant/android/issues/1574


dshokouhi (approved):
Like good to me, thank you for the update!

PR #539, opened 2021-06-07
Documentation for Local Push

zacwest:


TomBrien (approved):
From me this is good to go when you decide on local vs direct

PR #555, opened 2021-07-07
Clarify that android includes all data fields

dshokouhi:
Missed this bit earlier when we documented this feature. Thanks for the catch on this @zacwest :)


docs/notifications/actionable.md
```diff
@@ -186,6 +186,7 @@ When the notification action is performed, the `mobile_app_notification_action`
           "entity_id": "light.test",
           "my_custom_data": "foo_bar"
         }
+        // Android users can also expect to see all data fields sent with the notification in this response
```
zacwest, line 189: maybe include an example, like tag?

zacwest (approved):

PR #561, opened 2021-07-21
Some Hass.io -> Home Assistant updates

TomBrien:
Update references to Hass.io


docs/troubleshooting/networking.md
```diff
@@ -7,7 +7,7 @@ Having your Home available anywhere and everywhere you go is important, whether
...
 The easiest way is to subscribe to Nabu Casa Cloud for the small monthly fee of $5 US, which will solve all this for you and you can (almost) stop reading here, as well as supporting further development of Home Assistant. Nabu Casa Cloud acts as a "smart" proxy on the internet, tunnelling your frontend in an encrypted manner from your home to your phone, regardless where you are and without requiring opening your home network to inbound traffic from the internet.
 If you don't want to use Nabu Casa Cloud (which is fine, but you should still subscribe and enjoy the warm feeling of supporting Home Assistant), you need HA to be accessible from the internet. This requires opening a port on your router and getting a name for your Home Assistant on the internet. While it is possible to have your HA use Port 8123 internally and have your router do a port-forwarding from say the default https port of 443 to 8123, we recommend you NOT do this for reasons of simplicity which we will explain later. You also need a name for your Home Assistant as hassio.local is a private domain suffix that does not exist on the internet.
```
schmjop, line 13: Replace remaining occurence of hassio.local
```suggestion
If you don't want to use Nabu Casa Cloud (which is fine, but you should still subscribe and enjoy the warm feeling of supporting Home Assistant), you need HA to be accessible from the internet. This requires opening a port on your router and getting a name for your Home Assistant on the internet. While it is possible to have your HA use Port 8123 internally and have your router do a port-forwarding from say the default https port of 443 to 8123, we recommend you NOT do this for reasons of simplicity which we will explain later. You also need a name for your Home Assistant as homeassistant.local is a private domain suffix that does not exist on the internet.

```

schmjop (changes requested):
Replace remaining occurence of hassio.local

docs/troubleshooting/networking.md
TomBrien: Nice spot cheers

zacwest (approved):

frenck (approved):
LGTM, thanks 👍

PR #568, opened 2021-08-03
Clarification of Last Notification sensor

shieldmaiden95:
Added note to the last notification section clarifying that the sensor will not be created or available in Home Assistant until a notification is sent from one of the apps on allow list.


homeassistant:
Hi @shieldmaiden95, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#568).

Thanks!

dshokouhi:
Maybe it makes more sense to place this note at the top because there is more than 1 sensor that behaves like this.

shieldmaiden95:
> Maybe it makes more sense to place this note at the top because there is more than 1 sensor that behaves like this.

True, good point. Not sure how to update that now at this point. Looking into it now.

dshokouhi (approved):

Issue #232, opened 2020-05-27
Sunset migration guide


TomBrien:
We have loads of great iOS on-boarding info in the [migration guide](https://companion.home-assistant.io/docs/getting_started/migration). However, 2019.1 has been out for about 6 months now and I would hope most people have migrated by now. We should probably consider moving a lot of the problem solving steps to a troubleshooting set up page as the current title isn't clear that these steps apply to fresh installs too

PR #580, opened 2021-10-01
Remove migration guide

TomBrien:
Removes the migration guide from ios-1.5 to ios-2019.1 which has now been out for nearly 2 years and the guide contained some obsolete info. Adds appropriate redirect. Some minor auto linting sneaked in too. Fixes: #232


dshokouhi (approved):
Looks good to me

PR #586, opened 2021-10-07
Keep screen on option for webview activity

skynetua:


dshokouhi:
Could you add the beta flag to the new section please? This feature won't be available to all users until merged and a release is cut.

PR #592, opened 2021-10-13
Initial commit of Wear OS docs (beta)

dshokouhi:
We have now gotten some functionality that end users can enjoy so its time to get the beta version out soon. Support is very minimal. I wanted to get this page up and running to make it easier for contributors to document when they are adding more features.

I had a small change to the document structure section on the home page. Happy to change any text we see fit there.

Not sure what's the best way to organize the home page and side bar at the moment.

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -1,0 +1,12 @@
+---
+title: "Overview"
+id: "wear-os"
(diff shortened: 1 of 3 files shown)


dshokouhi:
Will hold off on merging until the listing is live, once its ready

PR #616, opened 2021-11-19
Update theming.md to include `primary-background-color` and `app-header-background-color` for android

LasseRosenow:
The android app uses `app-header-background-color` for its status bar coloring. See: [Ref](https://github.com/home-assistant/android/blob/672e49b6fbd48baff466f0a0ba7dfb3292768602/app/src/main/java/io/homeassistant/companion/android/webview/WebViewActivity.kt#L558)

And will soon be using `primary-background-color` for its navigation bar coloring. See: https://github.com/home-assistant/android/pull/1939

### DIFF
--- a/docs/integrations/theming.md
+++ b/docs/integrations/theming.md
@@ -5,15 +5,16 @@ id: 'theming'
 
 ## Colors used
 
-![Android](/assets/android.svg) &nbsp; Android<br />
-- `primary-color` for the status bar background color ![Android](/assets/android.svg)
+![Android](/assets/android.svg) &nbsp; Android<br /><br />
 Colors should be specified in hex format (e.g. `#0099ff`) and defining element colors through variable names is not supported.
+- `primary-background-color` for the navigation bar background color ![Android](/assets/android.svg)
+- `app-header-background-color` for the status bar background color ![Android](/assets/android.svg)
 
-![iOS](/assets/iOS.svg)<br />
+![iOS](/assets/iOS.svg)<br /><br />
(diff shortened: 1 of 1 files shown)


docs/integrations/theming.md
```diff
@@ -5,15 +5,16 @@ id: 'theming'
 
 ## Colors used
 
-![Android](/assets/android.svg) &nbsp; Android<br />
-- `primary-color` for the status bar background color ![Android](/assets/android.svg)
+![Android](/assets/android.svg) &nbsp; Android<br /><br />
```
dshokouhi, line 8: Since all these changes are currently in the beta lets add the beta flag here after the logo, then once we cut a release we will remove it

`<span class="beta">BETA</span><br />`

dshokouhi (changes requested):
I know what we have currently is outdated but lets just make sure to mark whats new as beta so users don't get confused :)

LasseRosenow:
You mean keeping the old as is and adding the new android documentation on top with the beta logo, so that users still have all the information?

dshokouhi:
> You mean keeping the old as is and adding the new android documentation on top with the beta logo, so that users still have all the information?

Sorry I was under the impression it was already outdated prior to your PRs, yes lets just keep the old information as is and have a new section for the beta changes.

LasseRosenow:
Sorry I was a bit unprecise. You are right, the old documentation is outdated. I changed `primary-color` from the old documentation to `app-header-background-color` and added `primary-background-color` in the beta documentation now

dshokouhi (approved):
Looks good to me, thank you for fixing the outdated info too!

PR #617, opened 2021-11-21
making 'how to add android sensors' a bit more clear to find

Viss:
added some plain-english folks would be searching/googling for in the pursuit of enabling the additional android sensors

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -62,6 +62,8 @@ Attributes such as `Cellular Technology` can be accessed with a template such as
 
 ![Android](/assets/android.svg) Android Sensor List
 
+How to enable additional sensors for android (like last_notification, for example):
+
 Each ![Android](/assets/android.svg) sensor below can be enabled by navigating to the `App Configuration` page then selecting `Manage Sensors`. By default, most are disabled with the exception of the [battery sensors](#battery-sensors) and any that were given permission during onboarding. Once enabled the sensor will begin to send data to your Home Assistant server, if you chose to disable it later on the sensor will stop updating. Upon enabling a sensor the app will request for permissions, if required. If you do not see a sensor listed below then your device does not support it. Some of the sensors below offer custom settings for each of their own needs, read about each one to see what it offers. These settings can be found in the same location where you enable the sensor.
 
 All sensors update during a periodic 15-minute interval and they will also update if other certain conditions are met. Read about each sensor below to understand how often to expect updates. During the 15-minute update interval a low priority foreground notification is temporarily created to prevent the Android system from halting the worker. This notification does not make a sound unless the user has installed a third-party app that intercepts notifications and decides to make a sound. If you are on Android 8.0+ you are free to minimize and/or turn off the notification channel for the `SensorWorker`.


homeassistant:
Hi @Viss, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#617).

Thanks!

docs/core/sensors.md
```diff
@@ -62,6 +62,8 @@ Attributes such as `Cellular Technology` can be accessed with a template such as
 
 ![Android](/assets/android.svg) Android Sensor List
 
+How to enable additional sensors for android (like last_notification, for example):
```
dshokouhi, line 65: The paragraph below actually talks about managing the sensor including disabling the sensor. In fact this sentence is similar to the one just underneath so it's not providing additional information.

dshokouhi (commented):
I don't see how this is any different than the sentence just below it. The entire paragraph is more than just enabling sensors.

docs/core/sensors.md
Viss: The intention here is to specifically use plain-english language - phrasing that people who would look for this stuff would find 'easy to find'. certain specific phrasing like "how to enable additional sensors" or, specifically "how to enable last_notification". Based on what I've seen in the github issues and the forums, folks will search for the specific thing that is ailing them, and work backwards from there, and currently the language in the sensors page, while articulate and bountiful, doesn't 'speak the same way' as a natural-langauge-google-search.

I spent two days looking while the answer was staring me right in the face - and my hope here is to have whoever the next person is not need to spend two days looking because the turn of phrase they were trying to search for wasnt present in the documentation.

It also doesnt help that the paragraph about android sensors is right below a table for IOS sensors, so to have the whole page sort of jump back and forth between IOS and android will lead folks to do what I did - presume the paragraph under the IOS table is still in the IOS section.

It may also be beneficial to have all of the android notes in one section, and all of the IOS notes in another section, so that its clear what goes where.

dshokouhi: >Based on what I've seen in the github issues and the forums, folks will search for the specific thing that is ailing them, and work backwards from there, and currently the language in the sensors page, while articulate and bountiful, doesn't 'speak the same way' as a natural-langauge-google-search.

You are the first user to get stuck on how to enable a sensor from any github issue posted.

> I spent two days looking while the answer was staring me right in the face - and my hope here is to have whoever the next person is not need to spend two days looking because the turn of phrase they were trying to search for wasnt present in the documentation.

> It also doesnt help that the paragraph about android sensors is right below a table for IOS sensors, so to have the whole page sort of jump back and forth between IOS and android will lead folks to do what I did - presume the paragraph under the IOS table is still in the IOS section.

If you actually read the page from top to bottom following the android icon this wouldnt be an issue. The docs are designed to keep iOS and Android together with distinguishing OS icons. The page isn't jumping here. The page starts with an explanation of what sensors are and what sensors are supported and then goes on to mention details about each sensor. That flow makes sense. 

If we are to add anything of value here we should change this text to be more generic because these sections are more about managing sensors rather than just enabling. We shoudl also make them an anchor link to become part of page navigation. But in all honesty if a user doesn't spend the time to properly read the docs they will have a tough time like you did.

dshokouhi:
In order to continue with this PR we also need the CLA signed

dshokouhi:
Closing this PR as the sensors page has changed a lot and may already satisfy the changes proposed here. If you still feel there should be an update please submit a new PR with the updated changes for review.

PR #626, opened 2021-12-05
Update notification commands for urlencoded extras

mvn23:
Related to home-assistant/android#1989

Add examples for urlencoded extras to notification commands.

The line breaks in the string given with the `group` parameter don't work well with yaml color coding on the website. Should this be made into a single long line or can we keep it like this? In other words: do we prefer shorter lines or proper color coding?

### DIFF
--- a/docs/notifications/commands.md
+++ b/docs/notifications/commands.md
@@ -198,9 +198,7 @@ automation:
           title: "nodomain.freeyourgadget.gadgetbridge.Q_PUSH_CONFIG"
           data:
             channel: "nodomain.freeyourgadget.gadgetbridge"
-            group: "EXTRA_CONFIG_JSON:%7B%22push%22%3A%7B%22set%22%3A%7B%22\
-              widgetCustom0._.config.upper_text%22%3A%22Hi%22%7D%7D%7D\
-              :urlencoded"
+            group: "EXTRA_CONFIG_JSON:%7B%22push%22%3A%7B%22set%22%3A%7B%22widgetCustom0._.config.upper_text%22%3A%22Hi%22%7D%7D%7D:urlencoded"
 ```
 
 ## Do Not Disturb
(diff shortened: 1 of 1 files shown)


docs/notifications/commands.md
```diff
@@ -184,6 +184,25 @@ automation:
...
+    action:
+      - service: notify.mobile_app_<your_device_id_here>
+        data:
+          message: "command_broadcast_intent"
+          title: "nodomain.freeyourgadget.gadgetbridge.Q_PUSH_CONFIG"
+          data:
+            channel: "nodomain.freeyourgadget.gadgetbridge"
+            group: "EXTRA_CONFIG_JSON:%7B%22push%22%3A%7B%22set%22%3A%7B%22\
+              widgetCustom0._.config.upper_text%22%3A%22Hi%22%7D%7D%7D\
+              :urlencoded"
```
dshokouhi, line 203:
```suggestion
            group: >
              "EXTRA_CONFIG_JSON:%7B%22push%22%3A%7B%22set%22%3A%7B%22\
              widgetCustom0._.config.upper_text%22%3A%22Hi%22%7D%7D%7D\
              :urlencoded"
```

This should render it properly :) May need to add some spacing I think though

mvn23: That does indeed get the formatting right, but it also breaks the string:
```yaml
group: >
  "string_a\
  string_b"
```
With the above, key `group` gets a value of `string_a\ string_b` instead of `string_astring_b`, which is what we're aiming for...

dshokouhi: oh good point :) yea we definitely don't want that then as the goal is for users to be able to copy and paste the examples to use them in automations. I think in the interest of formatting lets keep it all in one line as even in HA docs all the multi line examples shown to users follow the format I suggested which doesn't yield the correct results here.

mvn23: Done as suggested.

dshokouhi (approved):
Looks good to me, can be merged when the parent PR is merged

PR #627, opened 2021-12-07
Add Audio Output Sensors, rename Microphone to Audio Input

zacwest:


TomBrien (approved):

TomBrien:
This is a breaking change on the audio input side right?

zacwest:
Nope. Existing sensors will get a new visual name but keep the same entity ID.

PR #624, opened 2021-11-30
Change app configuration references

ludeeus:
For https://github.com/home-assistant/frontend/pull/10733

PR #635, opened 2021-12-13
Revert "Change app configuration references"

TomBrien:
Reverts home-assistant/companion.home-assistant#624 pending home-assistant/frontend#10890


ludeeus:
<home-assistant/frontend#10890> has changed and is now _only_ affecting non-admin users
Admin users still have to use Configuration -> Companion app

TomBrien:
Oh that's a pain to document!

TomBrien:
Given the current situation is a stop gap and the App Configuration will be more obvious for non-admin users than admin user. I'm going to close this rather than explain two different routes to app config everywhere

PR #639, opened 2021-12-20
Add location privacy note for multi-server in iOS-2022.2

zacwest:


TomBrien (approved):
This is a really excellent feature that is super true to HA's core concepts

PR #641, opened 2021-12-20
Remove old android note

dshokouhi:
This has not been true for a while....oops


dshokouhi:
closing in favor of #642

PR #643, opened 2021-12-21
Add sections to sensors page

zacwest:
Adding multi-server to this page gets more intense without sections for it, see #644.


dshokouhi (approved):
Long overdue :)

PR #650, opened 2022-01-06
Use local search instead of Algolia

TomBrien:
Switches from using Algolia/DocSearch for site searching to local search. Akin to home-assistant/developers.home-assistant#1176.


frenck (approved):
Thanks, @TomBrien 👍

PR #654, opened 2022-01-08
Add a page for documenting Quest specific features

dshokouhi:
As we just had the blog post done we should probably have a page ready to document Quest specific features and also to mention where to get the app and how it works.

Not sure if we should also put the link to the app on the home page?


TomBrien:
Yeah let's add a matching app store link to home and downloads pages

TomBrien:
Nice. Looks good on home page but for some reason is coming up a little too big on downloads. Also, sidequest is a brilliant name for a sideloading store on VR!

TomBrien (approved):
The download page sizing is a super minor thing. Feel free to merge this and get it out there then we can fox that down the line.

PR #655, opened 2022-01-08
Correct Description of Android Geolocate sensor

robclark56:
Acccording to this Google doc (https://developers.google.com/maps/documentation/geolocation/overview):
"The Geolocation API returns a location and accuracy radius based on information about cell towers and WiFi nodes that the mobile client can detect."

That matches the experience I see in what is reported with the HA Companion App installed on my Android phone. The reported data seems a lot less accurate than the GPS location I get on the same phone (using a GPS app)

iPhone: No idea. I do not have one.

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -92,7 +92,7 @@ All sensors update during a periodic 15-minute interval and they will also updat
 | `sensor.current_time_zone` | [See Below](#current-time-zone-sensor) | The current time zone the device is in. |
 | `sensor.current_version` | None | The current installed version of the application. |
 | `sensor.do_not_disturb` | None | The state of do not disturb on the device. |
-| `sensor.geocoded_location` | [See Below](#geocoded-location-sensor) | Calculated address based on GPS data. |
+| `sensor.geocoded_location` | [See Below](#geocoded-location-sensor) | Calculated address based on mobile phone cell towers and WiFi nodes that the mobile client can detect. |
 | `binary_sensor.high_accuracy_mode` | None | The state of high accuracy mode on the device. |
 | [Keyguard Sensors](#keyguard-sensors) | None | Sensors that represent various states about the device being locked or secured. |
 | [Mobile Data Sensors](#mobile-data-sensors) | None | Several different sensors around the state of mobile data. |


homeassistant:
Hi @robclark56, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#655).

Thanks!

dshokouhi:
This is technically not completely correct. The Android app uses Geocoded, not Geolocation. It uses the same data provided by the fused location service which is what the device tracker uses. So it can be cell tower, wifi and also GPS.

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt#L61

https://github.com/home-assistant/android/blob/master/app/src/full/java/io/homeassistant/companion/android/sensors/GeocodeSensorManager.kt#L78

Also see the link provided under how Android gets the location data, the one you linked is actually not used here as its a web service and not an android API: https://companion.home-assistant.io/docs/core/sensors#geocoded-location-sensor

TomBrien:
For apple we have much the same, all location updates are geocoded independent of source I believe

UBWH:
OK - I see that my companion app on Android sends these attributes as part of the phone sensor data:

- Location (for want of a better name, with Street Address, etc). We are told this is what Google provides via the [Geocoder](https://developer.android.com/reference/android/location/Geocoder) service; it can be pretty inaccurate.
- Accuracte GPS coords. 
- plus other phone data

While I have not checked the HA Companion code, but *IF* Geocoder is being passed correct GPS coords by the Companion  App, then Google is doing a poor job of converting that to an address. Using the same coords in maps.google,com is very accurate.

In summary:
- The Location (address) in the sensor data seems no more accurate than that from WiFi/Cell tower only.
- Either
  - Google Geocoder is doing a poor job, or
  - The Companion App is not giving Google Geocoder the correct GPS coords.

I am happy that the documentation accurately describes what the Companion app is trying to do (Ask Google to convert GPS -> Location address) - so happy to withdraw this pull request.

Before I do - any comments?

dshokouhi:
> but _IF_ Geocoder is being passed correct GPS coords by the Companion App, then Google is doing a poor job of converting that to an address. Using the same coords in maps.google,com is very accurate.

Look at the attributes of the geocoded sensor, you should be able to see the latitude and longitude used to obtain the geocoded location. Thats the same data we get from Google per the code above.

image


> * The Companion App is not giving Google Geocoder the correct GPS coords.

For this we request location data from google and then feed that into the geocoded API per the code link above.

UBWH:
Daniel,

100% agree. My phone is reporting very accurate Lat/Long figures. That is not the issue.

It is reporting poor Location (street address) data. Not consistent with the accurate GPS figures, but very consistant with  'locations' derived only from WiFi + Cell towers.

Just wondering why Google is not responding with more accurate street locations.

TomBrien:
The answer is probably Google doesn't know the location of your house super accurately. Are you get a close but address (say a neighbour) or totally wrong

dshokouhi:
Closing this PR as there is no real updates to push here.  We also link to the API being used if users require to know more about the specifics to how it works.

PR #680, opened 2022-02-05
Add button, input_button to Android QS tiles

jpelgrom:
Documentation PR for home-assistant/android#2247


jpelgrom:
Not sure how to integrate the BETA label here - perhaps it would be better to keep the current sentence, and just add a new paragraph before "The app currently offers 12 tiles ..."?

dshokouhi:
> Not sure how to integrate the BETA label here - perhaps it would be better to keep the current sentence, and just add a new paragraph before "The app currently offers 12 tiles ..."? (Or something similar to how the power menu/device controls documentation lists domains with the action, but that would be a bigger change.)

I think what you have currently is fine. Normally if I'm not sure I make a new paragraph and apply the beta label to that.

PR #694, opened 2022-02-22
Add a gallery of screenshots for the android app

dshokouhi:
I always felt the docs could benefit by having a gallery of screenshots showing all the screens that are created by the apps. Sometimes users may not know some features exist and sometimes its much easier to look at screenshots than to read text. Could be easy to reference, find more features etc...

Still need add some screenshots for Wear OS, will do in a separate PR to keep this one small

Also not sure if we should add screen shots for launcher shortcuts?

I dont think it makes sense to add screenshots for tiles and device control as the design on those are different from version to version and device to device.

@TomBrien  feel free to add screenshots for iOS/Mac if you find it might be helpful


TomBrien:
Looks good. Is it possible to tidy up the redaction on the discovery screenshot to be a square. 

I presume the idea here is a kind of "one-stop shop" for the images and describing the pages? Might be worth adding some low level sun headings to make linking easier. We should also make sure these are not better placed in appropriate guides, or placed there as well.

dshokouhi:
> Looks good. Is it possible to tidy up the redaction on the discovery screenshot to be a square.

done :) I probably shouldn't have done that on my phone directly 😝 



> I presume the idea here is a kind of "one-stop shop" for the images and describing the pages?

Kinda I just thought it would be nice to have a gallery of all the screens to help with discovery. Having 1 page to quickly scroll through will probably help with fatigue and discovery of new features.



> Might be worth adding some low level sun headings to make linking easier.

great suggestion, added more headings to make linking easier



> We should also make sure these are not better placed in appropriate guides, or placed there as well.

I think we can probably add these screenshots directly to the documented pages as well just to complete the look.

TomBrien (approved):
LGTM so feel free to merge, the only thing I'm wondering is if we should make this a new item on the top bar rather than in the docs

TomBrien:
I moved the gallery to the top menu, feel free to revert that commit if you're not a fan

dshokouhi:
> I moved the gallery to the top menu, feel free to revert that commit if you're not a fan

it looks a lot better this way actually, makes more sense too because its not docs just screenshots with brief descriptions :) Thank you for adding this commit!

Issue #701, opened 2022-03-02
Example for Actionable Notifications should mention that it blocks if notification is discarded


tamueller:
The example for actionable notifications here: https://companion.home-assistant.io/docs/notifications/actionable-notifications/ has a small caveat:

If the automation mode is set to Single, Queue or Parallel this will stop working once the notification is discarded in android (often enough). The trigger waits for one of the action, but if none of the actions is received, it will simply wait there, and block further executions of the automation.
There are two possible solutions for this: Either, briefly explain the problem and recommend using "restart" mode, or extend the example to also handle the "mobile_app_notification_cleared" event. I achieved this by simply adding:

`      - platform: event
        event_type: mobile_app_notification_cleared
        event_data:
          action_1_key: '{{ action_open }}'
`

I think the latter makes the example a little bit more helpful, but also more versatile, as the user immediately understands, that he can/should also handle the cleared event. Let me know if you want me to update the documentation or example-code.

dshokouhi:
> The trigger waits for one of the action, but if none of the actions is received, it will simply wait there, and block further executions of the automation.
> There are two possible solutions for this: Either, briefly explain the problem and recommend using "restart" mode, or extend the example to also handle the "mobile_app_notification_cleared" event. 

Actually you can also implement a `timeout` too so maybe it makes sense to have a considerations section?

https://www.home-assistant.io/docs/scripts/#wait-timeout

> I think the latter makes the example a little bit more helpful, but also more versatile, as the user immediately understands, that he can/should also handle the cleared event.

while it does make sense for android that means we will have to maintain 2 examples as the current example is compliant with iOS too

> Let me know if you want me to update the documentation or example-code.

Please feel free to do so, the docs only get better as we all come together to update them :)

tamueller:
@dshokouhi : Just a question, since I'm new to contributing to HA: Do I have to ask anyone directly to merge the pull request, or will this happen by itself if someone finds time?

dshokouhi:
> @dshokouhi : Just a question, since I'm new to contributing to HA: Do I have to ask anyone directly to merge the pull request, or will this happen by itself if someone finds time?

PR's will get reviewed when one of us finds the time, we get notified when they are submitted :)

PR #708, opened 2022-03-08
Updated actionable notifications documentation

tamueller:
Added further considerations on blocking behaviour
(as discussed in https://github.com/home-assistant/companion.home-assistant/issues/701)

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -240,24 +240,25 @@ When the notification action is performed, the `mobile_app_notification_action` 
 ### Blocking Behaviour
 The above example will wait, until the notification action is performed. This might lead to unexpected behaviour, depending on the [automation mode](https://www.home-assistant.io/docs/automation/modes/) of the script. For "single" mode, this will lead to a situation, where the script is not executed again if the previous notification action has not yet been performed. For "queue" and "parallel" this will happen if a certain number of notifications have not yet been performed. For "restart" mode it means, that as soon as the script is triggered again notification actions of the older instances of the script will not fire the coresponding action. Depending on the use case, there are several options:
 
--   In Android you can listen to the event that is fired when the notification is closed, and handle it accordingly. This can be achieved by adding the following lines
-```
-      - platform: event
-        event_type: mobile_app_notification_cleared
-        event_data:
-          action_1_key: '{{ action_open }}'
-```
-and 
-```
-    - conditions: "{{ wait.trigger.event.event_type == "mobile_app_notification_cleared" }}"
-      sequence:
-          - service: persistent_notification.create
-            data:
-              title: App notification result
-              message: The notification was closed
-```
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @tamueller, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#708).

Thanks!

docs/notifications/actionable.md
```diff
@@ -235,6 +235,31 @@ When the notification action is performed, the `mobile_app_notification_action`
 }
 ```
 
+## Further Considerations
+
+### Blocking Behaviour
+The above example will wait, until the notification action is performed. This might lead to unexpected behaviour, depending on the [automation mode](https://www.home-assistant.io/docs/automation/modes/) of the script. For "single" mode, this will lead to a situation, where the script is not executed again if the previous notification action has not yet been performed. For "queue" and "parallel" this will happen if a certain number of notifications have not yet been performed. For "restart" mode it means, that as soon as the script is triggered again notification actions of the older instances of the script will not fire the coresponding action. Depending on the use case, there are several options:
+
+-   In Android you can listen to the event that is fired when the notification is closed, and handle it accordingly. This can be achieved by adding the following lines
```
dshokouhi, line 243: Lets also link to the docs page here as it explains the cleared event in more detail: https://companion.home-assistant.io/docs/notifications/notification-cleared

docs/notifications/actionable.md
```diff
@@ -235,6 +235,31 @@ When the notification action is performed, the `mobile_app_notification_action`
...
+```
+    - conditions: "{{ wait.trigger.event.event_type == "mobile_app_notification_cleared" }}"
+      sequence:
+          - service: persistent_notification.create
+            data:
+              title: App notification result
+              message: The notification was closed
+```
+-   You can use a [time out](https://www.home-assistant.io/docs/scripts/#wait-timeout) to allow new executions of the script. However, this will lead to dangling notifications on your mobile phone. 
+-   It is possible to [clear notifications](https://companion.home-assistant.io/docs/notifications/notifications-basic#clearing) which can bei combined with timeouts and parallel execution mode to achieve good results.
```
dshokouhi, line 260: Typo here :)

dshokouhi (changes requested):
Just some minor comments to fix.

I do wonder though if we should place the last 2 options first because those options will work for iOS and Android while the first option is Android only.

tamueller:
@dshokouhi Hi Daniel, thanks for looking at the PR. I incorporated your comments and created another commit. As far as I understand github I do not need to create another PR, as the PR now contains both commits, right?

dshokouhi (approved):
Looks good to me, lets wait a day or 2 to see if the iOS team has anything to add :) Thanks for the update!

zacwest (approved):

PR #734, opened 2022-04-19
Apple Watch pages moved.

Benny-Git:
As discussed with @TomBrien in https://github.com/home-assistant/iOS/pull/2114 this should be fixed on the redirect.


TomBrien (approved):
Looks great, thanks for sorting this. And the app redirect is working again now (As you can see in the preview build here https://deploy-preview-734--home-assistant-companion-docs.netlify.app/app/ios/apple-watch)

PR #740, opened 2022-05-01
Update Android widgets docs for theming

jpelgrom:
Documentation PR for home-assistant/android#2496, and also adjusts the template widget formatting note with some examples because not all HTML tags will work.


dshokouhi (approved):

PR #748, opened 2022-05-14
Added documentation on how to install a certificate on the device whe…

magicbuffer:
Added documentation when Home Assistant requires TLS Client Authentication.

Relates to https://github.com/home-assistant/android/pull/2526


homeassistant:
Hi @magicbuffer, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#748).

Thanks!

dshokouhi (changes requested):
Let's add a beta flag with an android logo so users ware aware this is only available on the android beta app, once we cut a release the beta flag will be removed.

Also not sure if we should only document this on the troubleshooting page, might be hard to find. I think the Get Started page may also benefit if we add a android specific note since it seems to best fit there.

magicbuffer:
> Let's add a beta flag with an android logo so users ware aware this is only available on the android beta app, once we cut a release the beta flag will be removed.
> 
> Also not sure if we should only document this on the troubleshooting page, might be hard to find. I think the Get Started page may also benefit if we add a android specific note since it seems to best fit there.

It makes sense, I will move it to the getting started page and add the Android beta flag!

dshokouhi (approved):
Looks good to me, can be merged when parent PR is merged

PR #749, opened 2022-05-18
add info about multiple media players to media player widget docs

jannis3005:
This corresponds to PR [2534](https://github.com/home-assistant/android/pull/2534) in the homeassistant android repository.

### DIFF
--- a/docs/integrations/android-widgets.md
+++ b/docs/integrations/android-widgets.md
@@ -41,8 +41,8 @@ This widget will let the user control any media player on their home screen. The
 1.  Long press on any open space in the home screen
 2.  Scroll down to Home Assistant in the widget list
 3.  Drag the Media Player widget to an open space on the home screen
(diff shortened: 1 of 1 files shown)


dshokouhi (changes requested):
Sorry for the late review, this one fell off the radar. This feature is actually first available in the beta version of the android app. Would you be able to update this PR to split things up so users are aware the new features are only available in the beta with a beta tag? Once we cut a release we will remove the tags.

jannis3005:
I wasn't quiet sure on where/how to put them. is this as you intended @dshokouhi ?

jpelgrom:
The stable version of the app has actually been updated earlier this week so the beta tag is no longer necessary 🙂

dshokouhi:
My bad forgot to respond here and mention that :(

dshokouhi:
@jannis3005 any chance you can update this one more time? We don't need to mention beta anymore as this change is already in the production version.

dshokouhi (approved):

PR #751, opened 2022-05-20
fix notification url

glmnet:
Notifications seems to not need the `/lovelace` prefix anymore? Not sure about the actionable ones.

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -54,7 +54,7 @@ You can attach media and other content to notifications. See [Attachments](/docs
 
 When tapping on a notification, you can choose to open a URL, which can fall into one of the following buckets:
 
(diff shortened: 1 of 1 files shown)


dshokouhi (changes requested):
We should make sure to call out the HA core version where this changed as I think it was fairly recent (within past 6 months) otherwise users who did not update to this version will be confused.

glmnet:
Ja. Like meanwhile is better to have outdated docs. I'm done here

dshokouhi:
> Ja. Like meanwhile is better to have outdated docs. I'm done here 

The companion app docs always mention when certain versions are required. Not all users update immediately so it's best to be as informative as possible to be helpful to all users.

dshokouhi:
The example is actually valid before and after this change so lets leave it as is

PR #760, opened 2022-06-09
Add complication description

leroyboerefijn:
Accompanies https://github.com/home-assistant/android/pull/2574

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -50,6 +50,14 @@ Right now, two tiles are supported:
 * The shortcuts tile shows up to 7 shortcuts, which can be chosen from the settings section in the Wear OS app. You will be able to select the same set of entities you can access from the home screen.
 * The template tile shows a rendered template. The template can only be set from the android companion app. Note: it is not possible to scroll in a tile, the template should fit on the watch screen.
 
+## Complications
+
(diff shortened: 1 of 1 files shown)


docs/wear-os/wear-os.md
```diff
@@ -50,6 +50,14 @@ Right now, two tiles are supported:
 * The shortcuts tile shows up to 7 shortcuts, which can be chosen from the settings section in the Wear OS app. You will be able to select the same set of entities you can access from the home screen.
 * The template tile shows a rendered template. The template can only be set from the android companion app. Note: it is not possible to scroll in a tile, the template should fit on the watch screen.
 
+## Complications
+
+An entity state complication can be displayed on your watchface. The complication is of the 'short text' type and will display the current state of the selected entity. Depending on the watch face it will also show the entity name and icon. When you add an entity to a watch face, you're presented with a config view where you can select the entity to display. To change an entity, just change the complication and select the entity state complication again.
```
jpelgrom, line 55: > you're presented with a config view where

This is technical and can be removed. ("When you add an entity to a watch face, you can select the entity to display.")

> To change an entity

Not just any entity, but the selected entity.

leroyboerefijn: You're so quick haha. Thanks for the input!

jpelgrom:
The Wear OS page was updated because the app is also available without signing up for the beta, so this new section should now add a 'BETA' label (similar to what is done for changes in the main app).

PR #766, opened 2022-06-28
Remove troubleshooting step that is no longer valid

dshokouhi:
widgets no longer notify the user that they failed to get entity information, they now show the error directly on the widget instead of the toast message

the only remaining toast message is when a service call widget fails to send the service call so that one is still valid and the steps do not apply there.


jpelgrom (approved):
There are also still toast messages for the media player widget, but like the service button it's in response to user action and these troubleshooting steps do not apply.

PR #774, opened 2022-07-13
Add additional notification sensor attributes

dshokouhi:
Changes for: https://github.com/home-assistant/android/pull/2676

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -448,6 +448,9 @@ Below you can find some details that can be given with some notifications.
 | `is_ongoing` | If the notification is persistent on the device. |
 | `package` | The package that posted the notification. |
 | `post_time` | The time the notification was posted on the device. |
+| `channel_id` | The name of the channel the notification posted to. |
+| `group_id` | The name of the group the notification posted to. |
+| `category` | The category that the notification defined. |
 
 
 ### Media Session Sensor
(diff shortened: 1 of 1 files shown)


docs/core/sensors.md
```diff
@@ -448,6 +448,9 @@ Below you can find some details that can be given with some notifications.
 | `is_ongoing` | If the notification is persistent on the device. |
 | `package` | The package that posted the notification. |
 | `post_time` | The time the notification was posted on the device. |
+| `channel_id` | The name of the channel the notification posted to. |
```
jpelgrom, line 451: This isn't the name, but the ID of the channel. It might be the same (like for the HA companion), but doesn't have to be (it could be any string: English text even if the device is in another language, numbers, a UUID, ...).

Should there be a note that this will only show up on devices running Android 8 or newer?

docs/core/sensors.md
```diff
@@ -448,6 +448,9 @@ Below you can find some details that can be given with some notifications.
...
+| `group_id` | The name of the group the notification posted to. |
```
jpelgrom, line 452: This also is an ID instead of a name and could be anything (and I don't believe this one is visible anywhere for the user in the interface).

docs/core/sensors.md
dshokouhi: > This isn't the name, but the ID of the channel. It might be the same (like for the HA companion), but doesn't have to be (it could be any string: English text even if the device is in another language, numbers, a UUID, ...).

Code wise its almost the same with the exception that we do some "normalizing" in the code but yea I will update it to make it more correct :)



> Should there be a note that this will only show up on devices running Android 8 or newer?

Good call, will update it

docs/core/sensors.md
dshokouhi: I was probably thinking about the attributes a user defines when they send a notification to the app :)

JBassett (approved):

PR #781, opened 2022-07-24
Include CSP Frame Ancestors in android-webview.md

t-animal:
Include hint how to set up CSP Frame Ancestors directive in android webview documentation. Without this, it can be quite hard to figure out the origin to allow without remote debugging the webview.

### DIFF
--- a/docs/integrations/android-webview.md
+++ b/docs/integrations/android-webview.md
@@ -40,6 +40,9 @@ The ![Android](/assets/android.svg) Android app has the ability to enable Pinch-
 ## Remote Debugging
 The ![Android](/assets/android.svg) Android app has the ability to enable [chrome remote debugging](https://developer.chrome.com/docs/devtools/remote-debugging/) to allow for easier troubleshooting of front end issues. You can enable this setting in Companion App Configuration.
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @t-animal, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;companion.home-assistant#781).

Thanks!

jpelgrom:
Wouldn't this also apply to the iOS app, and make more sense on the troubleshooting page?

t-animal:
Possibly, I
But I don't know. I don't know exactly how the companion app and iOS webview works. I wouldn't move it there without confirmation, because no documentation is better than incorrect documentation.

jpelgrom:
@bgoncal Is this something you can verify for iOS?

bgoncal:
I didnt have time to validate that yet, but just looking at the docs it should be the same case for iOS https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors

Screenshot 2024-06-25 at 09 22 07

PR #796, opened 2022-08-20
Add note about Android template widget rate limits

jpelgrom:
Documentation PR for home-assistant/android#2801 based on a review comment.

Labeled with BETA because the app currently updates the template on any state change, so no rate limiting is applied (and it can be very inefficient).


dshokouhi (approved):
Looks good to me, can be merged when the parent is merged

PR #800, opened 2022-08-27
Fix volume level example after breaking change

dshokouhi:


jpelgrom (approved):
Oops, missed this in the previous review

PR #801, opened 2022-08-30
Fix rendering links for beta note on template widget

dshokouhi:
Noticed the links weren't rendering properly and looks like it needed some extra spacing :)

image


jpelgrom:
That's a weird bug

jpelgrom (approved):

PR #803, opened 2022-09-02
Switch from Oculus to Meta for the docs

dshokouhi:
Now that the official mobile apps have been updated I guess its time to update our docs 😄

### DIFF
--- a/docs/meta-quest/meta-quest.md
+++ b/docs/meta-quest/meta-quest.md
@@ -1,11 +1,11 @@
 ---
 title: "Overview"
-id: "oculus-quest"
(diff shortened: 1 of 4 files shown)


TomBrien (commented):
Do we want some redirects to handle people hitting to old oculus docs from direct links?

dshokouhi:
Oh right ya I guess we should add the redirect. Good call

dshokouhi:
@TomBrien did I get the redirect link correct here?

TomBrien (approved):
Yikes, sorry completely missed those. All good I think

PR #812, opened 2022-09-13
clarification for command_launch_app

adamaze:
Each time i try to use this feature, I forget how to find the "package_name". Hopefully this additional sentence in the doc will help me and others in the future. This was a relatively easy google on my part... but should we really expect people to google to find how to use "command_launch_app"?


dshokouhi:
This is a good suggestion but maybe we should improve upon the recommendation and show them an example using an app like Twitter?

Also I am not certain this is the best place to keep this as we have 3 other notification commands on the same page, notification [`clickAction`](https://companion.home-assistant.io/docs/notifications/notifications-basic#opening-a-url) and [`actions`]() as well as our [webview](https://companion.home-assistant.io/docs/integrations/android-webview#links) that depend on `package_name`

Maybe we should add this to the troubleshooting section and link to it from there on the other pages?

adamaze:
I would be happy to make that change. shall i include an actual link to the twitter app in the google play store?

dshokouhi:
Only if you feel that it is necessary, I actually had to guide a user on how to find the package name in the forums earlier so any example will be helpful to those that don't even know what a package name looks like :)

adamaze:
I'm looking at this again, and I'm honestly not sure how/where to add this... it would be silly to copy paste this one line into the other sections that use some form of *package_name... but then, do I make a new sub-section in the commands.md file explaining package_name, and then link to it from each section that uses *package_name?

jpelgrom:
You could add it to the general troubleshooting page and link to it from the documentation - it's not only useful for commands but also for e.g. actionable notifications.

For example [here](https://companion.home-assistant.io/docs/notifications/actionable-notifications/#android-android-specific) I would link "the package you wish to open" to the info and [here](https://companion.home-assistant.io/docs/notifications/notification-commands/#activity) I would link "provide the package" to the info, etc.

dshokouhi:
Closing as this PR has not moved and not many others have stumbled upon this. Lets revisit again if we feel the need and find a good central location.

PR #813, opened 2022-09-15
Update Networking - Setting up the app

jpelgrom:
Documentation PR for home-assistant/android#2883

I've updated the networking section app setup text to be more general / apply both to iOS and Android and fix a typo. A note has been added about the feature that's being added in the linked PR.


dshokouhi (approved):
Looks good to me, can be merged when the parent is merged.

PR #814, opened 2022-09-15
Docs for dynamic color sensor

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/2884

### DIFF
--- a/docs/core/index.md
+++ b/docs/core/index.md
@@ -591,6 +591,13 @@ Not all features are supported by Android at the moment but eventually most feat
       <td></td>
     </tr>
     <tr>
+      <td><a href="/docs/core/sensors#dynamic-color-sensor">Dynamic Color</a></td>
+      <td>✅</td>
+      <td>✅</td>
+      <td></td>
+      <td></td>
+    </tr>
+    <tr>
       <td><a href="/docs/core/sensors#pedometer-sensors">Floors Ascended</a></td>
       <td></td>
       <td></td>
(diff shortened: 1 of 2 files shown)


docs/core/index.md
```diff
@@ -590,6 +590,13 @@ Not all features are supported by Android at the moment but eventually most feat
       <td></td>
       <td></td>
     </tr>
+    <tr>
+      <td><a href="/docs/core/sensors#dynamic-color-sensor">Dynamic Color</a></td>
```
jpelgrom, line 594: This should match the sensor manager's name (Dynamic Color_s_), or the sensor manager should be renamed to match the documentation + Google

docs/core/sensors.md
```diff
@@ -101,6 +101,7 @@ You can change the frequency of sensor updates by navigating to Companion App in
 | `sensor.bluetooth_connection` | [See Below](#bluetooth-sensor) | The state of the sensor will reflect the total number of connected bluetooth devices. |
 | `sensor.current_time_zone` | [See Below](#current-time-zone-sensor) | The current time zone the device is in. |
 | `sensor.current_version` | None | The current installed version of the application. |
+| `Dynamic Colors` | RGB Color | The hexadecimal color value for the accent color used in the current device theme. <span class="beta">BETA</span> |
```
jpelgrom, line 104: This table lists:
- sensor managers, which are linked + don't use backticks, or
- individual sensors, which are not linked + use backticks
You're currently mentioning a sensor manager but are listing it like a sensor 😉

docs/core/index.md
dshokouhi: I wasnt sure if we should keep it plural in anticipation of adding more dynamic color sensors, what do you think?

dshokouhi: I suppose we could just keep it singular until we get a request for additional sensors, will update to reflect that :)

jpelgrom: Singular matches how it is described on material.io (which is linked here) 👍

jpelgrom (approved):

PR #816, opened 2022-09-17
Update URL handler page to account for Android support

dshokouhi:
Looks like we forgot to update this page to no longer be iOS specific after the android app added support for the navigation feature which I believe was done as part of the NFC addition.

Made some adjustments to make it more clear which parts are iOS only

### DIFF
--- a/docs/integrations/url-handler.md
+++ b/docs/integrations/url-handler.md
@@ -3,12 +3,14 @@ title: "URL Handler"
 id: 'url-handler'
 ---
 
-![iOS](/assets/iOS.svg)<br />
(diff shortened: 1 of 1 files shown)


jpelgrom:
"Navigate" mentions:

> This requires version 2021.3 or later.

Is this the core or app version? This should be clarified and potentially updated if it's the app version to also note the Android requirement.

dshokouhi:
Good question, I assumed at the time for HA core but it could've been an app version. In either case I wonder if we should call out versions from this long ago? 2021.3 was 18 months ago.

jpelgrom (approved):

PR #822, opened 2022-09-23
Update commands.md

chriskemp:
To match my proposed code changes in [PR # 2906.](https://github.com/home-assistant/android/pull/2906)

### DIFF
--- a/docs/notifications/commands.md
+++ b/docs/notifications/commands.md
@@ -384,7 +384,7 @@ If you have updated the Android app to 2022.8 you must use `command` in place of
 
 ![Android](/assets/android.svg)
 
-Users can turn the high accuracy mode of the background location sensor on or off using `message: command_high_accuracy_mode` with the `command` being either `turn_off`, `turn_on`, `force_off` or `force_on`. If `command` is blank, not set or not one of the above expected values then the notification will post as normal.
+Users can turn the high accuracy mode of the background location sensor on or off using `message: command_high_accuracy_mode` with the `command` being either `turn_off`, `turn_on`, `force_off` or `force_on`<span class='beta'> (the force options are BETA)</span>. If `command` is blank, not set or not one of the above expected values then the notification will post as normal.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


dshokouhi:
Thank you. As features get merged into the app they are actually part of the beta cycle. Could you please add the beta tags `<span class='beta'>BETA</span>` where appropriate so users know its part of the beta? Later on when a release gets merged we will remove the tags.

chriskemp:
Hope that's appropriate?

docs/notifications/commands.md
```diff
@@ -384,7 +384,7 @@ If you have updated the Android app to 2022.8 you must use `command` in place of
 
 ![Android](/assets/android.svg)
 
-Users can turn the high accuracy mode of the background location sensor on or off using `message: command_high_accuracy_mode` with the `command` being either `turn_off` or `turn_on`. If `command` is blank, not set or not one of the above expected values then the notification will post as normal.
+Users can turn the high accuracy mode of the background location sensor on or off using `message: command_high_accuracy_mode` with the `command` being either `turn_off`, `turn_on`, `force_off` or `force_on`<span class='beta'> (the force options are BETA)</span>. If `command` is blank, not set or not one of the above expected values then the notification will post as normal.
```
dshokouhi, line 387: I think for this change in particular it might best to mention the beta changes in its own highlighted section below the example as you have it only because it reads a bit odd here and may confuse some users. We will reincorporate the beta changes back into the description when the next release is cut.

docs/notifications/commands.md
```diff
@@ -401,6 +401,8 @@ automation:
             command: "turn_off"
 ```
 
+<span class='beta'>BETA </span>The difference between turn and force is only relevant if you have zone or bluetooth constraints set in the high accuracy mode settings.  In this case `force_on` will make high accuracy mode active until either `force_off` is sent, or the constraints go from active to inactive.  Similarly, `force_off` will turn off high accuracy mode until either `force_on` is sent, or the constraints go from inactive to active.
```
dshokouhi, line 404: The backtick code formatting does not look correct when I view the preview action link. I think you will need to add some new lines after the span tags so they render properly. You may also want to add one more new line at the end since the preview makes it look too close.

image

chriskemp:
Wierd - it looked ok in my browser preview - I don't get a red BETA either.  Anyway, I hope this is simpler.

dshokouhi (approved):

JBassett (approved):

PR #825, opened 2022-09-25
Fix incorrect statement on feature requests

frenck:
This PR adjusts the documentation of the companion app to fix an incorrect statement on where to suggest new features.

See also: https://www.home-assistant.io/help

We should not send users to all kinds of different places or create separate rules across the project org.

Feature requests in our issue trackers should be closed as a follow up.


TomBrien:
I have no problem with this, and an org wide approach makes sense. But to be clear at the moment the iOS repo new issue form points feature requests to Discussions in the same way that frontend does. Is this now heavily discouraged over a single point on the forums.

CC @home-assistant/ios-developers @home-assistant/android

jpelgrom:
Also agree that an org wide approach is good to have, but I fear feature requests might get lost / there will be a lot of duplicates if everything is moved to a single forum category. While the current approach isn't perfect, at least it's relatively easy to find feature requests for [something] once you know where to look.

dshokouhi:
Would it be possible to add categories in the Feature Request section so we can subscribe to specific android/iOS requests? Otherwise I am afraid a lot of the requests will get lost in the mix of things with HA core and Frontend etc...

I am all for this change as well.

PR #828, opened 2022-10-01
Update ble transmitter description for recent updates

dshokouhi:
Noticed that we were still mentioning enable all functionality even though it was removed a while ago. 

Also mentioning the new measured power setting as it was not mentioned before. 

Added links to the commands.

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -275,7 +275,7 @@ Settings are available to change the UUID, Major and Minor masks. These can be u
 There are also settings to alter:
 *   the Transmit power (between Ultra Low, Low, Medium and High)
 *   the Advertise mode (between Low Power (1Hz), Balanced (3Hz) and Low latency (10Hz))
-*   the Measured power (must be a negative number)
+*   the Measured power at 1 meter (must be a negative number)
 
 A Transmit setting toggle will start or stop the BLE transmissions. This setting as well as most of the above settings can be changed via the [notification command](../notifications/commands.md#ble-beacon-transmitter).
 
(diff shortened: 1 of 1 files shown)


docs/core/sensors.md
```diff
@@ -270,20 +270,21 @@ A BLE Transmitter sensor allows your device to transmit a BLE iBeacon.  This is
...
+Settings are available to change the UUID, Major and Minor masks. These can be used to change the overall identifier, as well as to allow groups, e.g. family phone devices can have particular Major value which can be whitelisted in apps like roomassistant. These settings are validated: UUID should be the [standard format](https://en.wikipedia.org/wiki/Universally_unique_identifier), Major and Minor need to be within 0 and 65535.
+
 There are also settings to alter:
 *   the Transmit power (between Ultra Low, Low, Medium and High)
 *   the Advertise mode (between Low Power (1Hz), Balanced (3Hz) and Low latency (10Hz))
-*   whether to allow this sensor to be turned on when the Enable all sensors toggle is activated (by default this is set to false, to prevent this sensor draining battery unnecessarily)
+*   the Measured power (must be a negative number)
```
jpelgrom, line 278: I still don't understand what this does, but the in-app description also mentions that this is "measured at 1m", should that be included here as well?

dshokouhi: I am actually trying to look at the repo and java docs to make sense of these variables to describe them a bit better here. Also just noticed we need to fix the description for the sensor since I see `#n` instead of `/n` 😂

jpelgrom: The answer was relatively easy to find with a quick search :) Looks like this value might be used by scanners for calibration and distance estimates. Knowing that the updated description that mentions it is at 1 meter distance is fine in my opinion.

jpelgrom (approved):

PR #833, opened 2022-10-05
Mention BLE transmitter works with iBeacon integration

dshokouhi:
Now that the iBeacon integration is live lets mention that the sensor works with it :)


jpelgrom:
As the docs are supposed to reflect the current state of Home Assistant, is "introduced in Home Assistant 2022.10" required? The minimum version is also mentioned on the linked page.

dshokouhi:
> As the docs are supposed to reflect the current state of Home Assistant, is "introduced in Home Assistant 2022.10" required? The minimum version is also mentioned on the linked page.

oh good point, didnt consider the requirement is mentioned there. Will remove :)

jpelgrom (approved):

PR #834, opened 2022-10-07
Add docs for wear OS network sensors

dshokouhi:
docs for: https://github.com/home-assistant/android/pull/2956

Note: I opted not to separate the mention of permissions as a beta feature as its more a general statement. The production version sensors will still enable as those don't require permissions. Adding in that note in a separate section may seem a bit odd, even if at the bottom.

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -72,4 +72,14 @@ Hint: use a [template sensor](https://www.home-assistant.io/integrations/templat
 
 ## Sensors
 
-The Wear OS app will have [sensors](../core/sensors.md). To start with, it will only report the [battery sensors](../core/sensors.md#battery-sensors). Its important to note that sensor updates require the app to post a notification to the device in order to prevent it from being killed by the OS. You can go to into Wear device settings and turn off the SensorWorker Notification channel to stop these notifications from buzzing on your wrist.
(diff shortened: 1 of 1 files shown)


jpelgrom:
> please refer to the link to learn more about how sensors update on Android

Is this correct for Wear OS? Considering how connectivity might be disabled when in standby and how complications update, I assume that WorkManager jobs might be significantly delayed to preserve battery life on watches, whereas on other Android device types it is fairly consistently every 15 minutes.

dshokouhi:
> > please refer to the link to learn more about how sensors update on Android
> 
> Is this correct for Wear OS? Considering how connectivity might be disabled when in standby and how complications update, I assume that WorkManager jobs might be significantly delayed to preserve battery life on watches, whereas on other Android device types it is fairly consistently every 15 minutes.

its going to be difficult to capture exactly how it works on all devices, I think the Pixel watch does its own thing too. The main thing to consider here is that the implementation is the same and intents should hopefully work with it. In my testing work manager resumes once connectivity is restored to send the update.

jpelgrom (commented):
Looks good to me!

In the future, if the number of sensors available on Wear OS increases significantly and/or there is more to document for other features it might be time to start splitting this up into separate pages.

jpelgrom (approved):

JBassett (approved):

PR #847, opened 2022-10-26
Change Wear OS Sensor list to a table

SkechyWolf:
Changes the Wear OS Sensor list to a table to match other sensor lists across the documentation website

### DIFF
--- a/docs/wear-os/sensors.md
+++ b/docs/wear-os/sensors.md
@@ -2,7 +2,6 @@
 title: "Sensors"
 id: 'sensors'
 ---
-![Android](/assets/android.svg)
 
 The Wear OS app also offers [sensors](../core/sensors.md#android-sensors) to consume your wearable data in Home Assistant, please refer to the link to learn more about how sensors update on Android. Not all sensors offered by the phone app will be offered by the Wear OS app. Please see the list below for what sensors are currently supported by the Wear OS app. If a sensor requires permissions you will be prompted to accept, otherwise the sensor will not enable and send data.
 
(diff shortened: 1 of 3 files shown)


dshokouhi:
@TomBrien  any chance you can help figure out why the build failed here? I also noticed the last PR merged did not update, wonder if its the same issue?

jpelgrom:
The build error should be fixed, @SkechyWolf can you try rebasing on the latest master?

jpelgrom:
The page is getting quite long... If the aim is to match the 'main' sensors table:
 - Wouldn't it be better to add beta tags at the end of the description instead, like in the other table? (It will look a bit odd right now but most of the time there won't be so many new sensors)
 - Why not split it up in to a separate 'Sensors' page in the Wear OS category? That will allow for adding detailed descriptions for sensors if required below the table.

dshokouhi:
Looking good so far, I think we should still mention sensors maybe in the beginning of the overview?

jpelgrom:
Now that everything is in one table, it should be sorted alphabetically in my opinion to make it easy to find the sensor you're looking for. So Battery moves down, and Bedtime mode moves up.

Do we need the Android logo at the top of the pages? The Apple Watch pages don't have an iOS/Apple logo, and you can also use the Wear app with an iPhone (if the watch supports it)...

SkechyWolf:
> Now that everything is in one table, it should be sorted alphabetically in my opinion to make it easy to find the sensor you're looking for. So Battery moves down, and Bedtime mode moves up.

Yeah does seem better to sort it alphabetically to find sensors easier

> Do we need the Android logo at the top of the pages? The Apple Watch pages don't have an iOS/Apple logo, and you can also use the Wear app with an iPhone (if the watch supports it)...

Yeah I was thinking of removing it after adding it because it seems redundant

jpelgrom:
Don't forget the Android logo on the overview page, other than that it looks good to me!

docs/wear-os/sensors.md
```diff
@@ -0,0 +1,37 @@
+---
+title: "Sensors"
+id: 'sensors'
+---
+
+The Wear OS app also offers [sensors](../core/sensors.md#android-sensors) to consume your wearable data in Home Assistant, please refer to the link to learn more about how sensors update on Android. Not all sensors offered by the phone app will be offered by the Wear OS app. Please see the list below for what sensors are currently supported by the Wear OS app. If a sensor requires permissions you will be prompted to accept, otherwise the sensor will not enable and send data.
+
+Its important to note that sensor updates require the app to post a notification to the device in order to prevent it from being killed by the OS. You can go to into Wear device settings and turn off the SensorWorker Notification channel to stop these notifications from buzzing on your wrist.
```
jpelgrom, line 8:
```suggestion
It's important to note that sensor updates require the app to post a notification to the device in order to prevent it from being killed by the OS. You can go to into Wear device settings and turn off the SensorWorker Notification channel to stop these notifications from buzzing on your wrist.
```

jpelgrom (approved):
Much easier to use now, thanks!

PR #850, opened 2022-10-27
Update yarn.lock to fix build with NodeJS 18

jpelgrom:
The LTS version for NodeJS changed from 16 to 18, which caused building the docs to fail with the error below. Regenerating `yarn.lock` / updating the dependencies' dependencies now while using 18 seems to fix this issue.

Similar issue/fix: home-assistant/developers.home-assistant#1514.

```
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
Error: error:0308010C:digital envelope routines::unsupported
    at String.replace (<anonymous>)
[ERROR] Client bundle compiled with errors therefore further build is impossible.
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,280 +2,153 @@
 # yarn lockfile v1
 
 
-"@algolia/autocomplete-core@1.7.1":
-  version "1.7.1"
-  resolved "https://registry.yarnpkg.com/@algolia/autocomplete-core/-/autocomplete-core-1.7.1.tgz#025538b8a9564a9f3dd5bcf8a236d6951c76c7d1"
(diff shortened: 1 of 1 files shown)


dshokouhi (approved):
it builds and previews, so looks good to me 😁

PR #865, opened 2022-11-27
Remove command bluetooth info

SamJongenelen:
Remove functionality from Android app


dshokouhi (changes requested):
This is not correct, the command is still valid for devices running Android 12 or older

Please see my response here: https://github.com/home-assistant/android/issues/3110#issuecomment-1328324378

SamJongenelen:
Yes, you are correct, thanks!

PR #867, opened 2022-11-29
Remove Android beta label for location tracking level

jpelgrom:
The location tracking level feature is included in the latest Android stable release, this label was missed.


dshokouhi (approved):
oops totally missed this one! thanks!

PR #870, opened 2022-12-17
Icon is mandatory in Quick Settings tile

stezak77:
Selecting an icon is a mandatory to create a Quick Settings tile.
Fixing documentation to avoid misunderstandings as user might think that if an icon is not set, the entity icon from HA server instance is automatically used, that is not the case.
Related to https://github.com/home-assistant/android/issues/3155


jpelgrom (approved):
Thanks!

PR #872, opened 2022-12-18
Doc Change: Clarified when notifications are clearable

thegareth:
Because of platform limitations, the application must be in memory to delete notifications.

Adding note to this section stating words to that effect.

https://community.home-assistant.io/t/clear-notification-not-working/319502/2

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -149,7 +149,7 @@ Clearing a notification on iOS requires app version 2021.5 or later.
 
 You can clear an existing notification which has a tag by sending `clear_notification`.
 
-Platform limitations may require the companion app to be have been recently used to clear the notification.
+Platform limitations may require the companion app to have been recently used to clear the notification: this applies for all iOS notifications, and any Android notifications not marked as critical.
 
 ![iOS](/assets/iOS.svg) will only clear the most recent [critical notification](critical.md) from a given tag.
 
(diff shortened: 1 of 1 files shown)


docs/notifications/basic.md
```diff
@@ -149,6 +149,8 @@ Clearing a notification on iOS requires app version 2021.5 or later.
 
 You can clear an existing notification which has a tag by sending `clear_notification`.
 
+Platform limitations may require the companion app to be have been recently used to clear the notification.
```
dshokouhi, line 152:
```suggestion
Platform limitations may require the companion app to have been recently used to clear the notification.
```

We may also want to make it clear this only applies to iOS. For Android as long as the notification is sent with the critical format we will wake up to clear it.

dshokouhi (changes requested):
Thanks for the contribution, just a small comment as this does not apply to android as long as the critical format was followed.

docs/notifications/basic.md
thegareth: Would this be better?

> Platform limitations may require the companion app to have been recently used to clear the notification: this applies for all iOS notifications, and any Android notifications not marked as critical.

dshokouhi: yes, thank you 🙏

dshokouhi (approved):

PR #884, opened 2023-01-13
Add docs for Wear OS notifications

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/3221

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -66,3 +66,28 @@ An entity state complication can be displayed on your watchface. The complicatio
 The complications are updated automatically whenever the screen is turned on and roughly every 15 minutes. You can force a complication to update by tapping it on the watch face.
 
 Hint: use a [template sensor](https://www.home-assistant.io/integrations/template/#state-based-template-binary-sensors-buttons-numbers-selects-and-sensors) for full flexibility.
+
+## Notifications
+
+<span class='beta'>BETA</span><br /><br />
+
+Wear OS devices will relay [notifications](../notifications/basic.md) from any phone app by default, meaning the notification needs to be sent to the phone first before it reaches the wearable. The Android app now allows for notifications to be sent directly to the watch itself, skipping the phone. Not all notification features supported by the phone app will be supported by the Wear OS app due to platform limitations.
+
+Currently only the following parameters are supported.
+
+*  `message`
+*  `title`
+
+Example:
+
+```yaml
+automation:
+  - alias: 'Send Wearable Notification'
+    trigger:
+      ...
+    action:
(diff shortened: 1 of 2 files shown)


jpelgrom:
While notification support on Wear is in development/very limited, should the notification page mention this/link to the Wear docs as well?

docs/wear-os/wear-os.md
```diff
@@ -66,3 +66,28 @@ An entity state complication can be displayed on your watchface. The complicatio
 The complications are updated automatically whenever the screen is turned on and roughly every 15 minutes. You can force a complication to update by tapping it on the watch face.
 
 Hint: use a [template sensor](https://www.home-assistant.io/integrations/template/#state-based-template-binary-sensors-buttons-numbers-selects-and-sensors) for full flexibility.
+
+## Notifications
+
+<span class='beta'>BETA</span><br /><br />
+
+Wear OS devices will relay [notifications](../notifications/basic.md) from any phone app by default, meaning the notification needs to be sent to the phone first before it reaches the wearable. The Android app now allows for notifications to be sent directly to the watch itself, skipping the phone. Not all notification features supported by the phone app will be supported by the Wear OS app due to platform limitations.
```
jpelgrom, line 74: > The Android app now allows for notifications to be sent directly to the watch itself

- The Android app: this could also refer to the app on the phone, which is incorrect
- _now_ allows: "now" isn't necessary as the docs are supposed to reflect the current state

> Not all notification features supported by the phone app

Phone app could be misleading as it is the same app on everything but a watch (tablets, Chromebooks, emulated on computers, ...), but I'm not sure how to refer to it (something like "main app" sounds too technical and suggests it is connected?) 🤔

dshokouhi: > Phone app could be misleading as it is the same app on everything but a watch (tablets, Chromebooks, emulated on computers, ...), but I'm not sure how to refer to it (something like "main app" sounds too technical and suggests it is connected?) 🤔

yea I suppose we could refer to it as the connected device but that does not make sense either in this scenario. How does google differentiate between it? I think theres just base android and wear OS restrictions?

jpelgrom: I can't find any user documentation from Google about limitations. Let's keep the updated text, in the future once more features are supported we can switch to marking unsupported features on the main notifications page instead of mentioning what is supported here.

dshokouhi:
> While notification support on Wear is in development/very limited, should the notification page mention this/link to the Wear docs as well?

Sorry missed this earlier, I added a note and link to the page :)

jpelgrom (approved):
Good to merge when the Android PR is merged

JBassett (approved):

PR #887, opened 2023-01-20
Initial Android Auto Docs

JBassett:


dshokouhi:
I think we should also add it to the landing page too so its more visible :)

https://github.com/home-assistant/companion.home-assistant/blob/master/src/pages/index.js

JBassett:
> I think we should also add it to the landing page too so its more visible :)
> 
> https://github.com/home-assistant/companion.home-assistant/blob/master/src/pages/index.js

Already Done. :-)

PR #888, opened 2023-01-20
Remove actionName from docs

jaamulberry:
Replace ```actionName``` with ```action```.

```actionName``` no longer works.

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -345,7 +345,7 @@ automation:
       - platform: event
         event_type: ios.notification_action_fired
         event_data:
-          actionName: "SOUND_ALARM"
+          action: "SOUND_ALARM"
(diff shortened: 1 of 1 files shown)


dshokouhi:
`actionName` seems to be used for a different event type altogether than `action`. That looks like an iOS specific event, lets wait for the iOS team to confirm if this is old and should be removed.

jaamulberry:
Hey @dshokouhi . That sounds good. If both are used the documentation should still be updated to call this out. Happy to add a commit with the language if the iOS team says it's still used. However I tested on Ipad and `actionName` did not work for me for an actionable notification

zacwest:
`ios.notification_action_fired` is a historic event, and the section you are modifying is "how to conver the historic event to the newer event"

PR #889, opened 2023-01-21
Add docs for Wear OS notification commands

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/3253

I also realized that a few additional parameters still apply to Wear OS and tested that they do indeed work so wanted to add these as well.

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -77,6 +77,8 @@ Currently only the following parameters are supported.
 
 *  `message`
 *  `title`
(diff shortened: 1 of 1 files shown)


dshokouhi:
Looks like we also create a `channel` however LED, vibration and even sound selection are not possible on Wear OS. I am not sure its worth mentioning it works if all the user can do is turn the channel on and off and these channels are set by the user with the exception of the `General` default channel.

JBassett (approved):

PR #901, opened 2023-01-31
Correct for iOS specific case

kloggy:
The sentence, "On success the sensor.last_update_trigger will change to "Push Notification" is iOS specific. Icon added.


dshokouhi (approved):
Thank you!

PR #902, opened 2023-02-01
Note that notification HTML formatting isn't always supported

jpelgrom:
Add a note to the docs indicating that HTML formatting in notifications is not supported on all devices / in all modes.

 - https://github.com/home-assistant/android/issues/2590
 - https://community.home-assistant.io/t/notification-message-html-formatting-color-text-doesnt-work-anymore/528531


dshokouhi (approved):
thank you!

PR #904, opened 2023-02-05
Add more supported wear OS notification parameters

dshokouhi:
Channel and vibration pattern do work, LED color may work but difficult if not impossible to test as I do not think any watches have a LED to control :)

~~Alarm stream may also work but is also not yet tested.~~ Alarm stream does not seem to work on Pixel Watch it most likely does not work on other watches.

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -87,6 +87,8 @@ Currently only the following parameters are supported.
 *  `title`
 *  [`notification_icon`](../notifications/basic.md#notification-status-bar-icon)
 *  [`tag`](../notifications/basic.md#replacing) (Support limited to replacing a notification only)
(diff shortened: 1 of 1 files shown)


dshokouhi:
Fixes: https://github.com/home-assistant/android/issues/3310 and https://github.com/home-assistant/android/issues/3323

jpelgrom:
What is the order for this list? When it was `message`, `title` and `tag` it made sense, now I don't see why for example `channel` is the fifth item in the list.

dshokouhi:
> What is the order for this list? When it was `message`, `title` and `tag` it made sense, now I don't see why for example `channel` is the fifth item in the list.

good call, its alphabetized now :)

jpelgrom (approved):

PR #909, opened 2023-02-13
Update complications.md

delprimero:
Minor grammatical edit


TomBrien (commented):
Sorry but I don't agree with this change. The sentence is saying that an icon can be selected in addition to setting the text. The suggested change seems to suggest that the template for the text is set by selecting an icon.

PR #914, opened 2023-02-21
Android per-server sensor management

jpelgrom:
Documentation PR for home-assistant/android#3357


dshokouhi:
Should add a note to both location and sensor updates that any settings applied will be used on all servers?

jpelgrom:
It's already in the docs, first line after the heading:

> (...) currently the sensor settings will be common for all connected servers.

dshokouhi (approved):

JBassett (approved):

PR #916, opened 2023-02-26
Add a note about beacon scanning foreground service

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/3369

### DIFF
--- a/docs/core/sensors.md
+++ b/docs/core/sensors.md
@@ -300,6 +300,10 @@ Settings are available to change scan period and interval which can be useful to
 
 A Monitor setting toggle will start or stop the scans - this setting can also be adjusted via the [notification command](../notifications/commands.md#beacon-monitor).
 
+<span class='beta'>BETA</span>
+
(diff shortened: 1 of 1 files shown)


jpelgrom:
> you will see (...) showing on the device

Isn't this the same but written differently?

> This is a foreground service to let the scan be more reliable in the background.

I think you can make this shorter and add it to the previous sentence without losing anything; "to improve background scanning reliability".

> This notification has a separate app provided channel which you can adjust on Android 8.0+.

This results in the user asking what the value of a notification channel is. Write that down instead, something like: "You can change the appearance of the notification on Android 8.0+ using the notification's channel."

jpelgrom:
> for the `Beacon Scanning`.

Wasn't it renamed to ['Beacon Monitor Scanning'](https://github.com/home-assistant/android/blob/194ed7ef2f62f8ae582cbbbc68de6e21be2a5fe3/common/src/main/res/values/strings.xml#L1047)?

dshokouhi:
> > for the `Beacon Scanning`.
> 
> Wasn't it renamed to ['Beacon Monitor Scanning'](https://github.com/home-assistant/android/blob/194ed7ef2f62f8ae582cbbbc68de6e21be2a5fe3/common/src/main/res/values/strings.xml#L1047)?

thats what I get for going by my outdated screenshot in the PR!

PR #926, opened 2023-03-12
Update attachments.md

Malnes:
just, why?!

### DIFF
--- a/docs/notifications/attachments.md
+++ b/docs/notifications/attachments.md
@@ -25,6 +25,10 @@ You can use relative URLs in the format `/media/local/direct.jpg` with this inte
 A file stored in `/media/file.jpg` on-disk is represented by `/media/local/file.jpg` in the notification. Note the addition of the `local` part of the path.
 :::
 
+:::warning
+An image stored in /media folder will be publicly available (no authentication required) after it has been sent as a notification attachment
+:::
+
 ### `www` folder
 
 You will want to store the image in the `www` folder located in your Home Assistant [config directory](https://www.home-assistant.io/docs/configuration/). This will then expose the image to the internet so that you may use it in your notifications and receive them anywhere.


TomBrien:
Can you please supply some context/background for this. The intended behaviour is that the app handles the authentication of the media folder as described [here](https://www.home-assistant.io/integrations/media_source/) and [here](https://companion.home-assistant.io/docs/notifications/notification-attachments#media_source-recommended)

If this is the case, a bug report should be raised, not a change to the docs.

Malnes:
I don’t see anything in the documentation explaining why files stored in /media suddenly becomes publicly available, but after much testing, that is the case.

First, I created a snapshot:

service: camera.snapshot
data:
  filename: /media/testImg.jpg
target:
  entity_id: camera.cam3_sd_stream



At this stage, the following url Returns 401, which is good.
https://<domain>.com/media/local/testImg.jpg
Then i added the image to a Notification:



service: notify.mobile_app_malnes_mobil
data:
  message: Test message
  title: Test message
  data:
    image: /media/local/testImg.jpg
    actions:
      - action: URI
        title: Open
        uri: https://<domain>.com/media/local/testImg.jpg<https://%3cdomain%3e.com/media/local/testImg.jpg>




After sending the Notification, the image is available for the world to see.
Don’t know if it’s a bug or a feature, you tell me.

Also, the image is available even after deleting the image from /media folder (getting some Eufy vibes here). Tried deleting and disabling cache from cloudflare, but that did not help either. It’s my understanding that HA use google notification services (firebase?) to deliver notifications, and I assume this is just a limitiation with that service.
OR this is just me doing something wrong…..



Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

zacwest:
Sounds like you have allowed Cloudflare to cache it somehow. I recommend looking there first; accessing HA directly will not allow access to the file without authentication. It makes sense that after validly pulling it your caching proxy could cache it.

Malnes:
I have disabled and deleted cache in cloudflare. I don’t think that’s the issue, but I will investigate further.


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

PR #928, opened 2023-03-17
Replace fix price by link to pricing table

c0ffeeca7:
- as the prices may differ slightly depending on the region.


dshokouhi (approved):

TomBrien (approved):
Nice spot, cheers!

PR #934, opened 2023-04-02
Add notes about android os sensors

frimtec:
Documentation for https://github.com/home-assistant/android/pull/3449


dshokouhi (changes requested):
Please revert the changes to the table in the sensor list so the change list only shows the added row instead of the whole table changing.

Please also add the beta tag to the table row and sensor description

dshokouhi (approved):

PR #939, opened 2023-04-20
Add clarification to the use of tags on persistent notifications

jack5mikemotown:
These changes update the documentation to make clear that while a tag must be defined, it does not have to be named "persistent". Based on the verbiage of "persistent" as the tag name, it may seem to some that the tag must have the specific name "persistent". To clear up any misconception, I changed the name of the tag from persistent to motion and clarified that the tag can be any value


dshokouhi (approved):
Thanks for your contribution!

PR #940, opened 2023-04-23
Error in example for TTS notification

genestealer:
sensor.<your_device_id_here>_is_music_active 
binary_sensor.<your_device_id_here>_is_music_active

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -544,7 +544,7 @@ automation:
         entity_id: sensor.<your_device_id_here>_audio_mode # Only speak if the phone is idle (not ringing or in a call)
         state: normal
       - condition: state
-        entity_id: binary_sensor.<your_device_id_here>_is_music_active # Only speak if the phone is not playing music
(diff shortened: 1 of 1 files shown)


docs/notifications/basic.md
```diff
@@ -544,7 +544,7 @@ automation:
         entity_id: sensor.<your_device_id_here>_audio_mode # Only speak if the phone is idle (not ringing or in a call)
         state: normal
       - condition: state
-        entity_id: sensor.<your_device_id_here>_is_music_active # Only speak if the phone is not playing music
+        entity_id: binary_sensor.<your_device_id_here>_is_music_active # Only speak if the phone is not playing music
```
dshokouhi, line 547: Good spot, its actually `binary_sensor.<your_device_id_here>_music_active` so we can also drop the `_is`

https://github.com/home-assistant/android/blob/master/common/src/main/java/io/homeassistant/companion/android/common/sensors/AudioSensorManager.kt#L52

jpelgrom:
```suggestion
        entity_id: binary_sensor.<your_device_id_here>_music_active # Only speak if the phone is not playing music
```

PR #946, opened 2023-06-06
Update local push settings path

dshokouhi:
minor update for new settings flow


jpelgrom:
Should we make it more like iOS? I don't like desired when most users will probably only have one server. "(tap on the server row/name)"

jpelgrom (approved):
I guess the commit means you agree :)

dshokouhi:
> I guess the commit means you agree :)

haha in the end we are saying the same thing, consistency is always good to have :)

PR #956, opened 2023-07-12
Add a note about persistent notification behavior starting in Android 14

dshokouhi:
We have a upcoming behavior change to one of our parameters so need to inform the user about this.

Ref: https://github.com/home-assistant/android/issues/3655

Actual docs: https://developer.android.com/about/versions/14/behavior-changes-all#non-dismissable-notifications

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -364,9 +364,6 @@ automation:
 
 ### Persistent Notification
 
-:::info
-Starting in [Android 14](https://developer.android.com/about/versions/14/behavior-changes-all#non-dismissable-notifications) persistent notifications will be dismissable by the user except when the phone is locked or user has selected "Clear All" button.
-:::
 
 Persistent notifications are notifications that cannot be dismissed by swiping away. These are useful if you have something important like an alarm being triggered. In order to use this property you must set the `tag` property as well. The `persistent` property only takes boolean (`true/false`) values, with `false` being the default. The persistent notification will still be dismissed once selected, to avoid this use `sticky: true` so the notification stays.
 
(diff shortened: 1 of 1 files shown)


jpelgrom:
Should the box be placed after the first paragraph perhaps, to allow users to understand what a persistent notification is first before detailing version-specific behavior? For example see the device compatiblity box after 'Notification Message HTML Formatting'.

> the user / user

Use 'you' like the main text.

> will be dismissable

Right now Android 14 is the future, but not once it is released; use 'can be dismissed'.

> the phone

More generic: the device.

dshokouhi:
thanks for the suggestion, applied the recommendations :)

jpelgrom:
You left a new blank line (367) ;)

jpelgrom (approved):

harmpter:
I'm very exited I found a solution via ADB
```
adb shell appops set --uid io.homeassistant.companion.android SYSTEM_EXEMPT_FROM_DISMISSIBLE_NOTIFICATIONS allow
```

I wonder whether it is appropriate to have the docs hint at this possibility. 
If yes, I would love to finally dip my toes into contributing to HA (the docs at least, lets start off cool right?)

mrdarrengriffin:
Oh that's really slick @harmpter

dshokouhi:
> I wonder whether it is appropriate to have the docs hint at this possibility.

forums would be a better place for unofficial fixes like this

jpelgrom:
> > I wonder whether it is appropriate to have the docs hint at this possibility.
> 
> forums would be a better place for unofficial fixes like this

Adding to that: including information like this in the documentation suggests (a certain level of) support for it by the HA developers, which is hard to provide for an internal flag in Android that may not work on every device or stop working.

PR #964, opened 2023-07-21
Add docs for default favorites view

dshokouhi:
docs for: https://github.com/home-assistant/android/pull/3694


jpelgrom:
I know this part hasn't changed, but for:

> If you would like to control only certain entities you can select some entities to be shown in the app.

the suggestion is that once you add a favorite, everything else disappears. That's not what happens. It's more about quick access to some entities when you open the app.

JBassett (approved):

PR #975, opened 2023-08-19
Use IPv6 prefix reserved for documentation

Tho85:
[RFC 3849](https://datatracker.ietf.org/doc/html/rfc3849) mandates to use the `2001:db8::/32` prefix for IPv6 documentation examples instead of randomly selected prefixes.


TomBrien:
Fair enough. Thanks!

TomBrien (approved):

PR #981, opened 2023-09-13
Update some android troubleshooting steps

dshokouhi:
Several updates:

- change TTS engine name to match latest app name for the package
- add a note about logging out and back in to start fresh steps as those are known to work
- mention to not login using trusted networks in troubleshooting and get started pages
- improve location troubleshooting steps to be more descriptive based on recent interactions with users


docs/troubleshooting/setup.md
```diff
@@ -106,6 +113,11 @@ If you are still seeing location issues then you may find it helpful to use the
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Begin evaluating if location update should be skipped
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Received location that is 74 milliseconds old, 1612371780829 compared to 1612371780903 with source fused
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Duplicate location received, not sending to HA
+```
+
+These are the logs you can expect to see when a duplicate location is received. The app will not send the same location update to the server if it has not changed for 15 minutes since the last update was sent.
```
jpelgrom, line 118: Shouldn't this be placed above the code block? At this point it is unclear if "these" refers to before or after. Above the code block there is already text about this first one, which would suggest this is about the second one.

jpelgrom:
I'm not sure about mentioning the trusted networks so prominently on the Getting started page. If we keep it there maybe move it out of the main text into a `:::info` block at the end?

docs/troubleshooting/setup.md
dshokouhi: ah yes I can rearrange this area

dshokouhi:
> I'm not sure about mentioning the trusted networks so prominently on the Getting started page. If we keep it there maybe move it out of the main text into a `:::info` block at the end?

as I have seen an uptick in users lately with this issue I opted to keep it and add it as a note like you suggested

docs/troubleshooting/setup.md
```diff
@@ -106,6 +113,11 @@ If you are still seeing location issues then you may find it helpful to use the
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Begin evaluating if location update should be skipped
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Received location that is 74 milliseconds old, 1612371780829 compared to 1612371780903 with source fused
 2021-02-03 09:03:00.903 7306-7306/? D/LocBroadcastReceiver: Duplicate location received, not sending to HA
+```
+
+Below you will find the expected logs for successful location results. If you do not see logs like this then make sure to follow the steps up above as more than likely the app does not have proper access to run in the background without any interference. If the Android system kills the app then you will not see these updates.
```
jpelgrom, line 118: While reading these sentences out loud you'd pause but there are not commas! Also trying to break it up to make it easier to read. 

```suggestion
Below you will find the expected log for successful location results. If you do not see lines like these, make sure to follow the previously mentioned steps. No location updates in the app log is usually caused by no proper access to run in the background without any interference, or because the Android system killed the app.
```

docs/troubleshooting/setup.md
```diff
@@ -116,7 +128,9 @@ If you are still seeing location issues then you may find it helpful to use the
 2021-02-03 09:06:34.309 7306-7430/? D/LocBroadcastReceiver: Location update sent successfully
 ```
 
-This is the expected logs for successful location results.  If you do not see logs like this then make sure to follow the steps up above as more than likely the app does not have proper access to run in the background without any interference.  If the android system kills the app then you will not see these updates.
+The logs will indicate whether a report was skipped due to time, accuracy or something else. If a report was skipped due to accuracy then double check the GPS coordinates to ensure they were correct and consider changing the [sensor setting for accuracy](../core/location.md#location-sensor-settings).
```
jpelgrom, line 131: Sometimes skipping a report due to accuracy is good, even if the location is roughly correct? This doesn't really tell you when you would want to change it (for example, because accuracy is 200 instead of 150, but for 1000 you'd still want to skip them).

jpelgrom: In that case, don't say "a report" but something like "multiple reports after another"? Sometimes you do get bad data, the accuracy shouldn't be changed just because you find one in the logs.

dshokouhi: ah yes we should be more explicit there :)

jpelgrom: I see you also added an example, thanks 👍

docs/troubleshooting/setup.md
```diff
@@ -116,7 +128,9 @@ If you are still seeing location issues then you may find it helpful to use the
...
 ```
 
-This is the expected logs for successful location results.  If you do not see logs like this then make sure to follow the steps up above as more than likely the app does not have proper access to run in the background without any interference.  If the android system kills the app then you will not see these updates.
+The logs will indicate whether a report was skipped due to time, accuracy or something else. If multiple reports are skipped due to accuracy then double check the GPS coordinates to ensure they were correct and consider increasing the [sensor setting for accuracy](../core/location.md#location-sensor-settings). If for example you see a valid report getting skipped with accuracy around `350` then set the setting to `400` as a buffer. Larger values may also lead to inconsistent results so go by valid reports in the logs.
+
+If you still experience an issue please submit a GitHub [issue](https://github.com/home-assistant/android/issues/new?assignees=&labels=bug&template=Bug_report.md&title=) and remember to attach the 10 minutes of logs from this troubleshooting step as its required to help.
```
jpelgrom, line 133: > If you still experience an issue

... after following these steps,

> as its required to help

Required is a bit harsh, and also not very privacy-friendly. To make it easier to help? If someone describes what's in the logs they aren't really required.

dshokouhi: good call updated to make it more clear and not required to help but more correctly "may be requested" :)

jpelgrom (approved):
Good to have, hopefully less confusing for users and better bug reports for us :)

PR #982, opened 2023-09-14
Android template widget text alignment example

jpelgrom:
Add an example to the docs for how to align text in a template widget [based on a recent interaction](https://github.com/home-assistant/android/issues/3864) + I've seen this question asked more often.


dshokouhi (approved):
Should we consider linking to source code here so users can get the full list of supported attributes?

jpelgrom:
The [Html.java soure code](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/Html.java) isn't really user-friendly with a list of supported tags/attributes or something like that, so I don't think it's useful. A lot of simple text related tags work as expected and the examples given should cover common use cases.

PR #992, opened 2023-10-01
Add Android NFC state sensor to the docs

marazmarci:
Added some basic docs for the new NFC state sensor added in https://github.com/home-assistant/android/pull/3905

(Also fixed some typos that I spotted along the way, please let me know if I should open a separate PR for those)


dshokouhi:
Can you please add the beta flag to the sensor in the list and it's description?

JBassett (approved):

PR #995, opened 2023-10-08
Update documentation about TLS CCA on Android

kleest:
This updates documentation about TLS Client Certificate Authentication changes in https://github.com/home-assistant/android/pull/3924.

### DIFF
--- a/docs/troubleshooting/networking.md
+++ b/docs/troubleshooting/networking.md
@@ -81,3 +81,5 @@ Note: If you don't use the NGINX Home Assistant add-on but instead roll your own
 
 If your Home Assistant requires TLS Client Authentication (because it is behind a reverse proxy configured to perform TLS Client Authentication), the app will ask for a certificate.
 Please refer to your device and Android version documentation to install the certificate, an example for Pixel phones is available here [Add & remove certificates](https://support.google.com/pixelphone/answer/2844832?hl=en).
+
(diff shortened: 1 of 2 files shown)


docs/troubleshooting/networking.md
```diff
@@ -81,3 +81,5 @@ Note: If you don't use the NGINX Home Assistant add-on but instead roll your own
 
 If your Home Assistant requires TLS Client Authentication (because it is behind a reverse proxy configured to perform TLS Client Authentication), the app will ask for a certificate.
 Please refer to your device and Android version documentation to install the certificate, an example for Pixel phones is available here [Add & remove certificates](https://support.google.com/pixelphone/answer/2844832?hl=en).
+
+Wear OS does not support authentication with installed certificates. The app cannot transfer the certificate to the Wear OS app automatically, therefore you are asked to provide a certificate during the Wear OS app onboarding. If a new certificate is required, you have to start the onboarding again by clearing the Wear OS app storage.
```
jpelgrom, line 85: Wouldn't logging out from the app, followed by logging in again be good enough as it sends new certificate data? (If not, that should probably be handled on logout as asking to clear all data shouldn't be required.)

jpelgrom (commented):
Could you add tags indicating this is feature requires a beta version of the app? See any other recent docs PR for an example.

docs/troubleshooting/networking.md
kleest: You are correct. I changed the wording.

kleest:
I added beta tags.

JBassett (approved):

PR #996, opened 2023-10-11
Update sensors.md

blakadder:
That line was not true since iBeacon integration filters out beacons without a name and doesn't discover them


dshokouhi:
Seems to be working for me

image

image

blakadder:
people disagree https://github.com/home-assistant/core/issues/80357

dshokouhi:
that issue will never get solved, too many users using a wide variety of different devices.

dshokouhi:
BTW device name is transmitted but its up to the device if it wants to transmit a name, typically it takes the BT device name but every android device here is different.

See this comment on some research done on this by the library that we use: https://github.com/home-assistant/android/pull/2941#issuecomment-1272379540

PR #997, opened 2023-10-11
Update sensors.md

blakadder:
how's about this as a solution then?


dshokouhi:
Are we certain users have adjusted the settings appropriately for their device? Its possible the settings are just not right because my devices do not get detected unless I change the settings.

blakadder:
I don't know about we but I'm not new at the testing game. I've tried all permutations of transmit options on multiple devices with three different instances of HA and 3 different BT receivers two of those being BT proxies. If a Beacon monitor from HA companion  detects the BLE transmitters from other devices I'd say the configuration is correct.

dshokouhi (approved):
Hopefully this helps settle the confusion.

PR #998, opened 2023-10-11
Remove beta tags for Android October 2023 release

dshokouhi:
We just cut a release so this should be merged once the release is live in the play store


jpelgrom (approved):

JBassett (approved):
We should probably consider using the next beta version rather than just a beta flag.  Then we don't have to cleanup after the release.

PR #1005, opened 2023-11-13
Added hotspot_state sensor.

RubenNL:
Documentation for a new feature.


dshokouhi (changes requested):
A couple things we need to do asides from adding to the list of network sensors.

- beta tag needs to be added as the feature will be part of beta until it goes live
- I dont think this sensor works on Wear OS does it? If not we need to mention that on the Wear OS docs

jpelgrom:
Why not in both places then? If you only add it on the main page it can be confusing for users who land on the Wear sensors page 
 🤷

It also should be "Wear OS", not "WearOS" :)

RubenNL:
The Wear OS sensor page has a link to the core page for the network sensors. Having to explain it there doesn't feel right to me.

Screenshot_20231113_185844

dshokouhi (approved):

JBassett (approved):

PR #1007, opened 2023-11-23
Document the UUID allowlist option

chatziko:
See:
- https://github.com/home-assistant/android/pull/4027
- https://github.com/home-assistant/core/pull/104419


dshokouhi:
Since this change will be in the beta first please add a beta flag so users know it's only available in the beta version one merged.

dshokouhi (changes requested):
Just need a beta flag to indicate its a beta feature/change initially

chatziko:
> Since this change will be in the beta first please add a beta flag so users know it's only available in the beta version one merged.

Done. I wasn't sure where to put it, to convey the fact that  BETA referes only to
>The iBeacon is capable of being detected by the [iBeacon integration](https://www.home-assistant.io/integrations/ibeacon) if its Minor field is set to the (default) value 40004

All the rest was valid before.

dshokouhi (approved):

jpelgrom:
@chatziko Is this PR still relevant? Looking at the core PR it seems that the approach is now to create a generic allowlist where the user could enter the app details, in which case this documentation change is incorrect and probably unnecessary.

chatziko:
> @chatziko Is this PR still relevant? Looking at the core PR it seems that the approach is now to create a generic allowlist where the user could enter the app details, in which case this documentation change is incorrect and probably unnecessary.

Indeed using a hardcoded allowlisted minor value was the initial approach ([old PR](https://github.com/home-assistant/core/pull/104419)), it was later replaced with a manual UUID allowlist ([new PR](https://github.com/home-assistant/core/pull/104790)). So this documentation change is no longer relevant. It should maybe be replaced by a sentence inviting users to allowlist their UUID.

dshokouhi:
@chatziko would you be able to update this PR to reflect the upcoming changes?

chatziko:
> @chatziko would you be able to update this PR to reflect the upcoming changes?

Done. Note that I removed the beta flag, since from what I understand it refers to the companion beta, but now the change is purely on the HA side.

PR #1011, opened 2023-12-08
Add hint how to create valid self-signed certificates

zi4k4j74yn:
Solution to https://github.com/home-assistant/android/issues/4041


dshokouhi:
I am a bit hesitant to accept this because we actually already do require that any certificate used must be valid

siegmund42:
Please have a look at the linked issue.
The challenge for a new user is to know which certificates are considered "valid" in android sense. Apparently the parameters from zi4k4j74yn's hint are essentially such that a self-signed cert is considered valid (after installing it as CA cert to android cert store).

I just verified the instructions work. When using 
`DNS:<your_dns_name_here>` instead of `IP:<your_ha_static_ip_here>` it even works using dns domain names instead of the static IP.

zi4k4j74yn:
I am fine with either merging or not merging, but I do think other users could profit from the information on how to create a certificate that is accepted.

siegmund42:
Actually, that was my first time running across this issue - most self
hosted applications I used before (nextcloud, davx5, etc) have an option to
allow self signed certificates. While this is not the most secure way (I
know), it is the easiest that most users are familiar with.
My guess is that, on the contrary, most users are unfamiliar with creating
a self signed certificate that is recognized as valid by Android OS.

Daniel Shokouhi ***@***.***> schrieb am Di., 16. Jan. 2024,
19:57:

hpbruns:
@zi4k4j74yn Thanks a lot for this hint. I am struggling a couple of weeks to bring my self signed cert running on my Android device. After using your tip, finaly ALL of my devices accept the certifacate.

PR #1012, opened 2023-12-10
Fix action name in actionable notification migration

myTselection:
align actionName and name of action to make it clear these should match to trigger on the desired event


jpelgrom:
Thanks for the suggestion, but it looks like the event name changed in step 1 from `SOUND_ALARM` (line 319) to `ALARM` (line 330). That means your suggestion updating it everywhere in step 2, specifically line 348, would still result in it not working as expected.

To fix this:
 - update line 319 to `ALARM`, OR
 - don't update line 348

(I wonder for how many people this migration info is still relevant 2.5+ years later)

PR #1024, opened 2024-01-28
Update siri-shortcuts.md for iOS 17

paull27:
Updated the shortcuts docs to reflect that in iOS 17 almost all shortcut triggers can run without confirmation.

The source for the new rules can be found here:
https://support.apple.com/en-ca/guide/shortcuts/apd602971e63/ios


SeanPM5:
Thanks @paull27

PR #1027, opened 2024-02-01
Drop pre-iOS 13 notes from actionable.md

adamjernst:
The companion app [no longer supports iOS versions before iOS 15](https://www.home-assistant.io/blog/2023/12/27/companion-app-for-ios-202312-lets-go/), so we can safely remove this.


dshokouhi:
isn't the app still available in the app store on lower iOS versions?

PR #1032, opened 2024-02-16
Update index.md

Emanabdalazez12:


dshokouhi:
What is the purpose of htis change? It looks like you added green checkmarks for everything. Table column headers do not need checkmarks, row headers do not need checkmarks. You are also checking items that the Android app does not support.

small example:
image

dshokouhi (changes requested):
Too many items with checkmarks please correct the checkmarks so only supported features have checkmarks on the appropriate platform.

PR #1039, opened 2024-03-03
Add Tile status entity

BlastyCZ:
I'm adding a new functionality (modification of the Android Tile) to the Home Assistant Android application. I want to control the garage door as quickly as possible. However, the switch that controls it does not correspond to the state but only opens or closes the garage door (it turns on for 1 second and then turns off). I have another entity for the state, but its state cannot be displayed. So, I added an option that allows the user to display (overwrite) the state, and I have only allowed binary_sensor.


jpelgrom:
Closing as parent PR is closed

PR #1042, opened 2024-03-10
Update supported Complication types within apple-watch.md

IT-Fuzzi:
Added a section to specify which Complication types are supported with watchOS 9.0 and higher.


bgoncal (approved):

T1ppes:
It seems that the Modular Watch Face has been removed by Apple and the previously Named Infograph Modular has been renamed to Modular. The Home Assistant Companion App still contains the references to the old watch face wording, which is confusing especially to those who do not have the background about the change done by Apple.

PR #1055, opened 2024-04-06
Update actionable.md

Didgeridrew:
Fixed competing quote mark issue in template condition in Actionable notification docs.


dshokouhi (approved):
Thanks for the fix!

PR #1057, opened 2024-04-16
Update networking.md

rubin110:
My apologies if this information should be provided on this page.

Adding in some helpful information about Let's Encrypt's plan to migrate away from using a 3rd party to cross sign their certs, and how that will impact users with older Android devices.

I've written a lengthier post here: https://community.home-assistant.io/t/lets-encrypts-ca-is-no-longer-considered-valid-on-android-versions-older-than-7-1-1/717907


docs/troubleshooting/networking.md
```diff
@@ -83,3 +83,5 @@ If your Home Assistant requires TLS Client Authentication (because it is behind
...
 
 Wear OS does not support authentication with installed certificates. The app cannot transfer the certificate to the Wear OS app automatically, therefore you are asked to provide a certificate during the Wear OS app onboarding. If a new certificate is required, you have to start the onboarding process again by logging out from the Wear OS app.
+
+Beginning February 2024 [Let's Encrypt has begun its migration over to providing keys that are not cross-signed by the IdenTrust's root CA certificate](https://letsencrypt.org/2023/07/10/cross-sign-expiration.html), instead using their own now widly trusted root CA cerficiate. Older clients that have not received an updated list of trusted certificate authorities will consider newer Let's Encrypt SSL keys as being invalid, which include versons of Android older than 7.1.1. This will cause Chrome or the Companion App to display an SSL error. An easy workaround is to use Firefox to access Home Assistant as it's distributed with its own updated trusted list of certificate authorities. Another work around is to manually install Let’s Encrypt's Active ISRG Root X1 Self-signed PEM (not DST Root X3) into the Android Credential Storage.
```
frenck, line 87:
```suggestion
Beginning February 2024 [Let's Encrypt has begun its migration over to providing keys that are not cross-signed by IdenTrust's root CA certificate](https://letsencrypt.org/2023/07/10/cross-sign-expiration.html), instead using their own now widely trusted root CA certificate. Older clients that have not received an updated list of trusted certificate authorities will consider newer Let's Encrypt SSL keys as being invalid, which include versions of Android older than 7.1.1. This will cause Chrome or the Companion App to display an SSL error. An easy workaround is to use Firefox to access Home Assistant, as it's distributed with its own updated trusted list of certificate authorities. Another workaround is to manually install Let’s Encrypt's Active ISRG Root X1 Self-signed PEM (not DST Root X3) into the Android Credential Storage.

```

frenck (approved):
Thanks, @rubin110 👍 

../Frenck

PR #1058, opened 2024-04-17
Add docs for remaining charge time sensor

dshokouhi:
docs for: https://github.com/home-assistant/android/pull/4346


jpelgrom:
This should mention the unit of measurement (ms)

dshokouhi:
> This should mention the unit of measurement (ms)

thanks for calling it out, updated to minutes per latest PR changes and also mentioning `unavailable` and `0` states as users may find `0` to be misleading when its a valid value from the API

PR #1069, opened 2024-06-18
Add certificate details for TLS Client Auth on Wear OS

DataGhost:
Using TLS Client certificates is not really straightforward going by the current docs and information in the app, resulting in either a single red X icon or a not-so-helpful "Could not register watch" message when incorrect or unsupported certificate files were used. This PR addresses at least the required format and, according to https://github.com/home-assistant/android/pull/3924#issuecomment-1784126322, also a possibly common error that might occur using the certificate container that otherwise works fine on the phone itself.


jpelgrom (approved):
Thanks for the addition. Looks like a good addition without turning it into technical step by step instructions to me.

PR #1071, opened 2024-06-22
Update camera widget documentation to include tap action configuration

michael-oknov:


dshokouhi (changes requested):
Thanks for the PR! As this will be a new feature it will first be part of the beta. Please add a beta label next to the features so users know that particular feature is part of the app beta. When we cut a release we will remove the beta labels. You may refer to a previous PR to see how beta labels are added

jpelgrom (approved):

PR #1073, opened 2024-07-03
Add update_widgets to the list of notification commands on iOS

literally-anything:
From: home-assistant/iOS#2830


bgoncal:
Thanks! I won't merge this one because I added it to this PR https://github.com/home-assistant/companion.home-assistant/pull/1081 that also adds more info

PR #1074, opened 2024-07-11
added command_activity example for VLC RTSP video stream

justin4480:
Adding a useful yaml example to trigger android table to open VLC media player and start an RTSP video stream.

### DIFF
--- a/docs/notifications/commands.md
+++ b/docs/notifications/commands.md
@@ -94,6 +94,24 @@ automation:
             intent_uri: "geo:0,0?q=1600+Amphitheatre+Parkway%2C+CA"
 ```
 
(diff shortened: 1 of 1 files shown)


dshokouhi:
This example does not seem to add anything other than being an example of how to use the feature with a certain app. The best place for that actually is to be in the forums. There is already a dedicated topic with lots of discussion going on regarding various apps. The examples here are meant to show how to work with new or different variables. I can already see in the forum topic some discussion regarding VLC and how users configured it.

https://community.home-assistant.io/t/android-intents-sending-receiving-list-starting-activities-too/276192

jpelgrom:
I'm going to agree with @dshokouhi and close this PR. The example adds no new information other than how to use VLC in this specific scenario. We could also add 100 other examples like it. The forums are the right place for this.

PR #1076, opened 2024-07-15
Add note about same group and tag for android

dshokouhi:
Fixes: https://github.com/home-assistant/android/issues/4453 because unexpected behavior can occur if a user changes the tag and group unexpectedly.

### DIFF
--- a/docs/notifications/basic.md
+++ b/docs/notifications/basic.md
@@ -136,6 +136,10 @@ Replace an existing notification by using a tag for the notification. All subseq
 
 ![iOS](/assets/iOS.svg) does not support replacing [critical notifications](critical.md).
 
+:::info
+![Android](/assets/android.svg) if you plan on using `tag` and `group` in the same notification then use the same value to avoid unexpected behavior.
+:::
+
 ```yaml
 automation:
   - alias: "Notify of Motion iOS replacement"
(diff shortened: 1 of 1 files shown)


jpelgrom:
I think this suggests both values need to be the same, which isn't true. The issue is probably the _repeated use with only one value changing_.

dshokouhi:
does that make it more clear?

jpelgrom:
Yes I think it is :)

docs/notifications/basic.md
```diff
@@ -136,6 +136,10 @@ Replace an existing notification by using a tag for the notification. All subseq
 
 ![iOS](/assets/iOS.svg) does not support replacing [critical notifications](critical.md).
 
+:::info
+![Android](/assets/android.svg) do not use the same `tag` in different `group` to avoid unexpected behavior.
```
jpelgrom, line 140:
```suggestion
![Android](/assets/android.svg) Do not use the same `tag` in different `group` to avoid unexpected behavior.
```

dshokouhi:
> Looking at the netlify preview: should we place the info block below the yaml perhaps, or remove the info block?

hmm good point, i am not sure we have a lot of info blocks here now 🙈 we do need to notify users to avoid confusion

docs/notifications/basic.md
```diff
@@ -150,6 +150,10 @@ automation:
             tag: "backyard-motion-detected"
 ```
 
+:::info
+![Android](/assets/android.svg) Do not use the same `tag` in different `group` to avoid unexpected behavior.
```
jpelgrom, line 154:
```suggestion
![Android](/assets/android.svg) Do not use the same `tag` in a different `group` to avoid unexpected behavior.
```

jpelgrom (approved):

PR #1082, opened 2024-07-25
fix missing macOS table header

darthsebulba04:
The table header is missing for macOS in the Notifications group.  This puts it in there so the table is complete.


dshokouhi (approved):
Thanks for the fix 🙏

PR #1086, opened 2024-08-01
Add docs for updating ble transmitter measured power

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/4546

### DIFF
--- a/docs/notifications/commands.md
+++ b/docs/notifications/commands.md
@@ -230,6 +230,24 @@ automation:
             ble_uuid: "b4306bba-0e3a-44df-9518-dc74284e8214"
 ```
 
+<span class='beta'>BETA</span>
+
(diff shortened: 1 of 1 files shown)


jpelgrom:
This section is getting a lot of examples, I'm wondering if perhaps adding things to a list/table like above would be better. Did you consider this?

dshokouhi:
> This section is getting a lot of examples, I'm wondering if perhaps adding things to a list/table like above would be better. Did you consider this?

yea a table does not make much sense here because we can only update 1 item at a time, only measured power can be updated and a separate command needs to be sent to turn off teh transmitter or adjust the advertise mode etc...

jpelgrom:
I suppose that is true. Let's save the discussion about the number of examples in the documentation for another time ;)

jpelgrom (approved):
Good to merge when Android PR is merged

PR #1091, opened 2024-08-07
Update service (call) to (perform) action

jpelgrom:
In line with the core 2024.8 release, rename `service:` to `action:` in yaml examples, and service calls to actions and executing calls to performing actions in text.

This PR does not update mentions of services in the app UI, as the app UI should be updated first.


dshokouhi (approved):
Looks good to me

frenck (approved):
Nice, thanks, @jpelgrom 👍 

../Frenck

PR #1094, opened 2024-08-12
Docs for additional android phone sensors

dshokouhi:
docs for: https://github.com/home-assistant/android/pull/4575


dshokouhi:
> I think you need to update the feature overview table to change phone state to phone sensors now.

ah good catch! updated!

jpelgrom (approved):

PR #1103, opened 2024-09-14
Add docs for navigation started event

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/4637


jpelgrom:
Maybe mention that the data contains the entity that was tapped? Core doesn't have "navigation entities" so this may not be clear.

dshokouhi:
> Maybe mention that the data contains the entity that was tapped? Core doesn't have "navigation entities" so this may not be clear.

good call, updated to be more precise :)

jpelgrom:
Keep it consistent: Event data will contain the selected entity ID.

PR #1105, opened 2024-09-16
Add note about person entity configuration

dshokouhi:
Find myself helping people a lot with this. I did not include it as a step because teh steps are for the `device_tracker` entity which is actually more important for the app as that is the only entity we update.


jpelgrom:
I feel like this should be a general FAQ, not Android specific, because you can mix and match Android and iOS and other OS phones/tablets/trackers ("Is the location for your person entity not updating as expected? Go to Settings > People and make sure that there are no devices linked to the person which always stay in the same place, like a tablet at home. or something like that) After adding that, we can link to it from this specific FAQ.

dshokouhi:
closing in favor of https://github.com/home-assistant/companion.home-assistant/pull/1112

PR #1110, opened 2024-09-24
Update sounds.md

VE6SRV:
Removed an extra "data:" tag in the example yaml code.

### DIFF
--- a/docs/notifications/sounds.md
+++ b/docs/notifications/sounds.md
@@ -23,7 +23,6 @@ automation:
     - action: notify.mobile_app_<your_device_id_here>
       data:
         message: “Your Roommate arrived”
-        data:
           push:
             sound: "US-EN-Morgan-Freeman-Roommate-Is-Arriving.wav"
 ```


dshokouhi:
only `message` and `title` are supported under the first `data` block I thought, pretty sure `push` needs to be under the second `data` block.

Looking at the FCM repo that seems to still be the case?

https://github.com/home-assistant/mobile-apps-fcm-push/blob/main/functions/ios.js#L72

VE6SRV:
I'm brand new at this, don't know much about yaml at all... however, having the second data: tag in there seemed to cause a problem. The sound notification didn't work. Studio Code Server would throw an error when I tried to copy exactly what was in the example.

However, I just pulled up the yaml in Studio Code Server, and that data tag is in there again, and the sound notification works.

Here I thought I found an issue, but it probably isn't an issue. This is a steep learning curve, I usually have to reference about 10 help pages to try and figure out how to do what I want to do.

The yaml that is working is below.

  - action: notify.mobile_app_myphone
    data:
      message: Turning on the outside and entryway lights.
      title: Welcome Home!
      data:
        push:
          sound: US-EN-Alexa-Welcome-Home.wav

Looks like the UI fixed what I edited, adding back in the data: tag that seemed to be my problem.

dshokouhi:
its quite possible you had a weird conflict in the structure, YAML is very precise and is easy to be finnicky. The example in teh docs followed the expected format. Glad to hear its working for you properly without any changes :)

VE6SRV:
Computers are very picky... that's the best part about dealing with them, you need to be very precise in what you tell them to do. Also the annoying part if you don't know what to tell them.

Should I close this, or delete this fork?

dshokouhi:
we can leave it closed and you are free to delete the fork if you like :)

VE6SRV:
I'll kill it. Thanks for the quick sanity check!

PR #1119, opened 2024-10-11
Add docs for car speed sensor

dshokouhi:
Docs for: https://github.com/home-assistant/android/pull/4722


jpelgrom:
Looks like you forgot to update the ID when copying the line :) I think you should either abbreviate (m/s) or write it out (meters per second), not a mix of both.

dshokouhi:
oops good catch 🙈 updated :)

jpelgrom (approved):
Can be merged when parent is ready

PR #1123, opened 2024-10-18
Docs update for Android October release

dshokouhi:
We have a production release pending approval from Google.

Also had to remove Health Connect at the last minute so also removing that for now, we should add those back after approval from Google.


jpelgrom (approved):
Release is live 🎉

Don't forget to add Health Connect back later

dshokouhi:
> Don't forget to add Health Connect back later

definitely but at this point best to wait until its approved

PR #1138, opened 2024-11-13
Update android-shortcuts.md

icebaer12:
Extended description for adding shortcuts to self-created dashboards.


docs/integrations/android-shortcuts.md
```diff
@@ -7,7 +7,7 @@ id: 'android-shortcuts'
...
+There are currently 2 supported shortcut types: Dashboard or Entity. The default shortcut type is Dashboard and you must enter the Dashboard path (ex: `/lovelace/default_view` or `/lovelace-dashboardname/viewname`) in order to create the shortcut. It is also possible to specify a direct URL as it is displayed in the browser (e.g. `/dashboard-mushroom/0` if your own dashboard is called 'Mushroom'). If a specific subpage (e.g. kitchen) of the dashboard 'Mushroom' is to be addressed, `/dashboard-mushroom/kitchen` can be entered. If you selected Entity then you will be presented with a new Entity field that will contain a list of all entities from your Home Assistant server to select. Once selected you will be able to create the shortcut.
```
dshokouhi, line 10: While this does provide an extra example mentioning direct URL may be confusing still. The app actually accepts the relative URL which we should call out here. Not sure we need to provide examples of each and every type of relative URL one can have but we can add to the list of current examples if need be. The concept of relative URLs is used in a few different places already in the docs.

https://companion.home-assistant.io/docs/notifications/notifications-basic/#opening-a-url 
https://companion.home-assistant.io/docs/notifications/actionable-notifications#uri-values
https://companion.home-assistant.io/docs/notifications/notification-commands#webview
https://companion.home-assistant.io/docs/integrations/url-handler#navigate

PR #1142, opened 2024-11-30
Add note about Google review for Health Connect sensors

dshokouhi:


jpelgrom:
Maybe you should change 495 to `:::note` so the added info stands out more? They are now similar 'weight' while the existing box feels less important.

jpelgrom (approved):

PR #1150, opened 2024-12-21
Add android app launching links examples

GiorgioAresu:
The documentation wasn't clear on how to dial a phone number (or mailto, etc.) on Android, it seemed it was only possible on iOS. I found how to do it and think it could be useful to somebody else

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -157,6 +157,13 @@ You can send a specific [deep link](https://developer.android.com/training/app-l
   uri: "deep-link://example://link/to/content"
 ```
 
+You can also use this for application-launching URLs. For example, to make a telephone call:
+```yaml
+- action: "URI"
+  title: "Deep Link"
+  uri: "deep-link://tel:2125551212"
+```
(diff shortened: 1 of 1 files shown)


docs/notifications/actionable.md
```diff
@@ -157,6 +157,13 @@ You can send a specific [deep link](https://developer.android.com/training/app-l
   uri: "deep-link://example://link/to/content"
```
dshokouhi, line 157: Lets keep this simple and just replace the current deep link example to show the `tel` example this way we cover both scenarios

dshokouhi (changes requested):
Lets reduce the amount of examples and use the current one to be more specific.

ideally the android should update to work with iOS example for the `tel` link but that would require code changes so making the current example represent what is live makes the most sense

docs/notifications/actionable.md
GiorgioAresu: Done. Are you happy with the wording?

dshokouhi (approved):

PR #1151, opened 2024-12-21
Update meta-quest.md

slaygirlz:
looks like the quest 3s wasn't added 👍

### DIFF
--- a/docs/meta-quest/meta-quest.md
+++ b/docs/meta-quest/meta-quest.md
@@ -18,7 +18,7 @@ Home Assistant has started to offer the [minimal flavor](/core/android-flavors.m
 | Meta Quest | Yes |
 | Meta Quest 2 | Yes |
 | Meta Quest Pro | Yes |
-| Meta Quest 3 | Yes |
+| Meta Quest 3/3s | Yes |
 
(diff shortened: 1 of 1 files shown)


docs/meta-quest/meta-quest.md
```diff
@@ -18,7 +18,7 @@ Home Assistant has started to offer the [minimal flavor](/core/android-flavors.m
 | Meta Quest | Yes |
 | Meta Quest 2 | Yes |
 | Meta Quest Pro | Yes |
-| Meta Quest 3 | Yes |
+| Meta Quest 3/3s | Yes |
```
jpelgrom, line 21: Let's match Meta's branding

```suggestion
| Meta Quest 3/3S | Yes |
```

jpelgrom (changes requested):
See my previous comment

slaygirlz:
sorry! I thought that maintainers can override pulls but here is the updated one

jpelgrom:
> sorry! I thought that maintainers can override pulls but here is the updated one

Yes, but you could've disagreed and it's not urgent :)

jpelgrom (approved):
Thanks!

PR #1155, opened 2025-01-05
Add thermostat tile to Wear OS documentation.

Martreides:
Added thermostat tile documentation.

### DIFF
--- a/docs/wear-os/wear-os.md
+++ b/docs/wear-os/wear-os.md
@@ -78,7 +78,7 @@ You may use HTML to format the text displayed. The following tags are currently 
 * Using headers: `<h1>title</h1>`, `<h2>subtitle</h2>`, etc.
 
 ### Thermostat tile
-<span class='beta'>The thermostat tile can be used to change the target temperature of a climate enitity up or down by the step size. Since tile refreshing is limited it can happen that the target temperature shown is no longer accurate. Therefor, if a user changes the temperature through the tile the current target temperature is always first refreshed. If a user changes the temperature multiple times quickly, each click changes the temperature relative to the result previous click instead of getting the target temperature from the server. This is done to counteract the delay in changing target temperature by some thermostats.</span>
+<span class='beta'>BETA</span> The thermostat tile can be used to change the target temperature of a climate entity up or down by the step size. Since tile refreshing is limited it can happen that the target temperature shown is no longer accurate. Therefor, if a user changes the temperature through the tile the current target temperature is always first refreshed. If a user changes the temperature multiple times quickly, each click changes the temperature relative to the result previous click instead of getting the target temperature from the server. This is done to counteract the delay in changing target temperature by some thermostats.
 
 ## Complications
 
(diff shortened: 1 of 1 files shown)


dshokouhi:
Thanks for the PR! Please add a beta label to the thermostat tile mentions so users are aware this is only available in the beta. Feature are first part of the apps beta cycle before moving to a prod release. You may refer to a previous PR for the beta label format.

Martreides:
Okay! how do I add a beta label?

dshokouhi:
> Okay! how do I add a beta label?

you add `<span class='beta'>BETA</span>` on the bullet point line item and also under the header for the tile description

dshokouhi:
So with the latest change all the text is now in red. The purpose of the beta tag was actually to just show the words "BETA" in red so users understand the next sentence/section is for the beta :) You can use the exact format above in the 2 places so it can say beta before the text.

docs/wear-os/wear-os.md
```diff
@@ -53,9 +53,10 @@ The settings screen can be found at the bottom of the home screen. This is where
 * The Template tile shows a rendered template. The template can only be set from the Android companion app. Note: it is not possible to scroll in a tile, the template should fit on the watch screen.
 * The Camera tile shows a snapshot of the selected camera.
 * The Assist tile allows you to quickly [open Assist on your watch](https://www.home-assistant.io/voice_control/android/#assist-on-wear-os).
-
+* <span class='beta'>The Thermostat tile allows you to see and control the target temperature of a climate entity.</span>
```
dshokouhi, line 56:
```suggestion
* <span class='beta'>BETA</span> The Thermostat tile allows you to see and control the target temperature of a climate entity.
```

docs/wear-os/wear-os.md
```diff
@@ -76,6 +77,9 @@ You may use HTML to format the text displayed. The following tags are currently
 * Changing the color: `<font color='#03a9f4'>colored text</font>`
 * Using headers: `<h1>title</h1>`, `<h2>subtitle</h2>`, etc.
 
+### Thermostat tile
+<span class='beta'>The thermostat tile can be used to change the target temperature of a climate enitity up or down by the step size. Since tile refreshing is limited it can happen that the target temperature shown is no longer accurate. Therefor, if a user changes the temperature through the tile the current target temperature is always first refreshed. If a user changes the temperature multiple times quickly, each click changes the temperature relative to the result previous click instead of getting the target temperature from the server. This is done to counteract the delay in changing target temperature by some thermostats.</span>
```
dshokouhi, line 81:
```suggestion
<span class='beta'>BETA</span> The thermostat tile can be used to change the target temperature of a climate entity up or down by the step size. Since tile refreshing is limited it can happen that the target temperature shown is no longer accurate. Therefor, if a user changes the temperature through the tile the current target temperature is always first refreshed. If a user changes the temperature multiple times quickly, each click changes the temperature relative to the result previous click instead of getting the target temperature from the server. This is done to counteract the delay in changing target temperature by some thermostats.
```

Martreides:
Thanks for the suggestions @dshokouhi, kinda forgot about the documentation :sweat_smile:

dshokouhi (approved):

PR #1157, opened 2025-01-10
Update health connect for recently approved sensors

dshokouhi:
Beta was just approved! Hopefully it stays that way.


jpelgrom (approved):
Release to production was approved

PR #1158, opened 2025-01-10
Update docs with corrected dashboard URLs

travissanderson:
Trying to follow the documentation here didn't end up navigating the app to the correct URL. I tried removing lovelace and it worked, so I updated the wording here accordingly. I also added an example of a subview since that may not be obvious to all readers.

### DIFF
--- a/docs/integrations/url-handler.md
+++ b/docs/integrations/url-handler.md
@@ -12,9 +12,12 @@ If multiple servers are connected to an app, you will be prompted to select a se
 :::
 
 ## Navigate
-This allows you to update the frontend page location via a deeplink.
+This allows you to update the frontend page location via a deeplink. To build a deeplink, follow these steps:
 
-For example: if you had a dashboard at `/webcams` you can use `homeassistant://navigate/webcams` to launch the app there. Similarly, you can navigate to a subview named `example` under your `/webcams` dashboard using `homeassistant://navigate/webcams/example`.
+1. Navigate to the link you'd like to deeplink to in the web app, e.g. `http://homeassistant.local:8123/dashboard-mobile/my-subview`
+2. Copy the path portion of the URL, in this example that would be `/dashboard-mobile/my-subview`
+3. Craft your URL by starting with `homeassistant://navigate` and adding the path, e.g. `homeassistant://navigate/dashboard-mobile/my-subview`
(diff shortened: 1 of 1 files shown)


dshokouhi:
`/lovelace/` is still a valid path for some like myself :)

image


This PR tries to solve a similar problem like #1154 

Maybe a better fix for this is to create a FAQ item for how to discover a users relative URL and link that to all the pages where relative URLs are used because more than 1 feature relies on it.

dshokouhi:
> Interesting, do you know why that is? I would guess new(er) users will not have lovelace and it is a legacy thing, so having this in the documentation would be better than the current. I suppose the more correct documentation may be to concatenate `homeassisant://navigate` and the path of the URL you see in the browser when looking at your dashboard in the web app?

yea have had HA for many years and i manage my main dashboard in YAML so definitely legacy on my end :)

i think the best course of action would be to not mention any specific format (as those may change later with new dashboards) and just guide the user to finding their relative path

travissanderson:
@dshokouhi let me know what you think of acccf04

dshokouhi:
Looks good but one exception the android app does not yet support step 4 :)

travissanderson:
@dshokouhi do you know if there's a github issue or PR for adding that I could link to? or maybe I can just comment on the planned work to update these docs

dshokouhi:
> @dshokouhi do you know if there's a github issue or PR for adding that I could link to? or maybe I can just comment on the planned work to update these docs

No PR or issue created for it, we can keep this as is and update the docs as needed later. We dont generally mention if things are requested or in progress. Not all features are supported by the other platform so just important to call out what platforms support the features.

travissanderson:
Okay, after looking at the netlify preview I realized we're now repeating the same information about 3 times regarding servers. I eliminated the step 4 I had added in favor of the subsection. Does the iOS icon indicate that all these deeplinks are only supported on iOS and Android can only do navigate?

travissanderson:
I think this page may need updated too then as it shows no URL Handler support for either Android minimal or full https://companion.home-assistant.io/docs/core/#feature-comparison

dshokouhi (approved):
Thanks for the improvement!

dshokouhi:
> I think this page may need updated too then as it shows no URL Handler support for either Android minimal or full https://companion.home-assistant.io/docs/core/#feature-comparison

ah good call out, yes that should be updated :)

PR #1188, opened 2025-03-30
Add doc entry for new todo list widget

kamilwawrzyczek:
Docs for https://github.com/home-assistant/android/pull/5169


dshokouhi (changes requested):
Thanks for the contribution and also including the beta tag! Just need to line up with how the frontend and existing documentation uses the hyphen. The description may also need to be tweaked again based on PR feedback.

https://www.home-assistant.io/integrations/todo/

dshokouhi (approved):

PR #1190, opened 2025-04-04
Android: Add details around proposed structure split for device controls

SheaSmith:
This PR adds details of a new option that would allow people with multiple servers to split their servers into different 'structures' (aka pages) within Android's device controls area. This is the documentation PR companion for home-assistant/android#5190


jpelgrom:
Could you add a beta label (`<span class='beta'>BETA</span>`) to the page, to indicate the feature requires the user to use a beta version of the app for now? Documentation updates are published at the same time as code is merged, and we'll remove the beta label when a new stable release is pushed.

dshokouhi (approved):

PR #1197, opened 2025-04-24
change wording on android version requirement

harmpter:
Android 12 or older -> Android 12 or below

## Proposed change

Hi, this is my first pull request and I hope I'm doing it correctly. 
The docs point out that `command_bluetooth` only works on android versions **Android 12 or older**, which is easily confused for Android 12+ in a rush. I would like to change the word "older" to "below", as I find it more straightforward and prevents misunderstanding of the fact, that it only works up until Android 12, not from Android 12+ and above.

## Type of change

- [ ] Document new or changing features for which there is an existing pull request elsewhere
- [x] Spelling or grammatical corrections, or rewording for improved clarity

## Checklist

- [x] I have read and followed the [documentation guidelines](https://developers.home-assistant.io/docs/documenting/general-style-guide).
- [x] I have verified that my changes render correctly in the documentation.

## Additional information

- This PR fixes or closes issue: Is related to [/android#3171](https://github.com/home-assistant/android/issues/3171) and [/android#3110](https://github.com/home-assistant/android/issues/3110)
- Link to relevant existing code or pull request:

### DIFF
--- a/docs/notifications/commands.md
+++ b/docs/notifications/commands.md
@@ -270,7 +270,7 @@ automation:
 
 ## Bluetooth
 
-![Android](/assets/android.svg) &nbsp;Android 12 or older
+![Android](/assets/android.svg) &nbsp;Android 12 or below
 
 Users can turn Bluetooth on or off using `message: command_bluetooth` with the `command` being either `turn_off` or `turn_on`. If `command` is blank, not set or not one of the above expected values then the notification will post as normal.
 


jpelgrom:
Thanks for the suggestion.

However, the documentation consistently uses "or older" for this kind of requirement, and in the text it is still a 5 letter word so I don't really see how this is less easily confused.

harmpter:
Thank you! I see, I wasn't aware of that. 
Regarding the semantics; while "or older" is definitely not incorrect, it is easy to make the mistake of thinking of it as "greater number = greater age = older" because that's how age usually works. 
I think "or below" doesn't carry that same ambiguity. An alternative would be "or earlier" with the added bonus of being a 7 letter word haha.
But I now recognize and agree that the issues were created around the time Android 13 rolled out, that's fair and I missed that.
But, we have no way to estimate how many other users have fallen into the same trap as I have, because the existing issues effectively prevent duplicate issues from being created. So, I would still argue to make **some** kind of change.

Can I add another commit to this pull to at least keep the wording consistent across the documentation? If yes, I will try to do that. Please let me know what you think about the older/below/earlier thingy.

jpelgrom:
Looking at the style guide recently adopted for Home Assistant documentation for [earlier](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/earlier), it looks like that's the way to go:

> Use _or earlier_ or _previous_ instead of _or lower_ to refer to all versions of a product that precede a particular release if the statement is accurate for all preceding releases.

There are a lot of places where "or older" is currently used - so if you want to update it, please update it everywhere for consistency.

PR #1257, opened 2025-12-01
Enhance connection security documentation with FAQ

bgoncal:
Added FAQ section addressing common questions about connection security options and VPN usage.

## Type of change

- [ ] Document new or changing features for which there is an existing pull request elsewhere
- [ ] Spelling or grammatical corrections, or rewording for improved clarity

## Checklist

- [ ] I have read and followed the [documentation guidelines](https://developers.home-assistant.io/docs/documenting/general-style-guide).
- [ ] I have verified that my changes render correctly in the documentation.

## Additional information

- This PR fixes or closes issue: Fixes #
- Link to relevant existing code or pull request:

### DIFF
--- a/docs/getting_started/connection_security_level.md
+++ b/docs/getting_started/connection_security_level.md
@@ -96,4 +96,4 @@ Go to **Settings** > **Companion app** > **[your server]** > **Connection securi
 
 Without location permission, the app cannot determine which network you're currently connected to. If you choose "Most secure" without granting location access, the app will only be able to use your external (HTTPS) URL.
 
-If you need to use an unencrypted internal URL and don't want to grant location permission, you'll need to select "Less secure"—but be aware of the security implications.
+If you need to use an unencrypted internal URL and don't want to grant location permission, you'll need to select "Less secure"—but [be aware of the security implications](#option-2-less-secure).


docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
 The **Most secure** option prevents these risks by ensuring unencrypted connections only work when you're on your home network. The **Less secure** option removes this protection, and is not recommended for most users.
 
 For maximum security, we recommend using HTTPS connections with valid SSL certificates always, especially when accessing Home Assistant remotely.
+
+# FAQ
+
+## If my internal and external URLs are already HTTPS, which option should I choose?
+
+In that case you don't even need 2 URLs in the App, you can use just one (set to external URL).
```
jpelgrom: Internal/external URL may still be different for advanced use cases so "you don't even need 2 URLs" isn't necessarily true.

Why is app capitalized?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+## Does this feature work when I'm on my home network through a VPN?
+
+No. On iOS, if you access your Home Assistant remotely via VPN, you need to:
```
jpelgrom: What doesn't work exactly - it's unclear what "this feature" refers to without reading the text? What about Android?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+1. Set your local IP as the **external URL**
+2. Make sure VPN is connected when you want to execute actions (e.g., running a script from a widget)
+
+## Should I enable this feature if I use a VPN?
```
jpelgrom: Shouldn't this be placed before the previous question?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+## If my internal and external URLs are already HTTPS, which option should I choose?
+
+In that case you don't even need 2 URLs in the App, you can use just one (set to external URL).
+
+## Does this feature work when I'm on my home network through a VPN?
+
+No. On iOS, if you access your Home Assistant remotely via VPN, you need to:
+
+1. Set your local IP as the **external URL**
+2. Make sure VPN is connected when you want to execute actions (e.g., running a script from a widget)
```
jpelgrom: I think these sentences are long enough that they should end with a `.` (https://learn.microsoft.com/en-us/style-guide/scannable-content/lists#punctuation). Are these alternatives or do you need to do both?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+## Should I enable this feature if I use a VPN?
+
+On **iOS** If you've set your local IP as the external URL, it doesn't matter whether you enable this feature or not, what is set as external will be used regardless.
```
jpelgrom: What about Android?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+**Important security consideration:** This setup is not advisable because if you're not connected to your VPN and happen to connect to a public WiFi, it could expose your credentials.
```
jpelgrom: Consider using a `:::warning`box for this instead to make it stand out more

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,27 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+## Will the app send credentials if there's no connectivity to my instance?
+
+Only use your local IP or any other non-https URL as internal URL, and configure your home network so the App knows when it's the right time to connect locally, if your local IP is set as the external URL it's considered a secure connection (since you shouldn't have non-HTTPS remote access). The app will behave as normal, for example, if you try to run a script from a widget, it will use your credentials to execute that script.
```
jpelgrom: This does not answer the question, and has a very long sentence that is hard to follow at the start. I tried writing a suggestion, but don't know whether the question is wrong or what the answer is trying to say.

docs/getting_started/connection_security_level.md
bgoncal: Good point, I was thinking on "same URL", but I didn't wrote that

docs/getting_started/connection_security_level.md
bgoncal: The page is regarding the connection security level feature.
Android will have to update this document when it is ready to publish, for now it is focused on iOS, which is what is in prod, but if you already have something in mind for Android, feel free to add

jpelgrom: If the page is focused on iOS right now, I wouldn't call out "on iOS" every time so we can add/update it when ready for Android if needed.

docs/getting_started/connection_security_level.md
bgoncal: Both

docs/getting_started/connection_security_level.md
bgoncal: Android specific content needs to be added later to this doc when the feature is in prod, for now I'm focusing on iOS

docs/getting_started/connection_security_level.md
bgoncal: If I remove "on iOS" and this doc is not updated on time, it will cause confusion, I prefer keeping iOS mentioned there and remove or tweak later when we also have the android side

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
 
 For maximum security, we recommend using HTTPS connections with valid SSL certificates always, especially when accessing Home Assistant remotely.
+
+# FAQ
+
+## Why am I seeing the connection security prompt?
+
+Starting with version 2025.11, the companion app will ask you to choose a security level if your setup includes an unencrypted URL (HTTP instead of HTTPS). This prompt helps protect your Home Assistant credentials from being exposed on public networks.
+
+If you have HTTPS for all your URLs and still saw this prompt, please update to the latest app version—a patch was released to address this.
```
jpelgrom: Apps don't get patche, and simply telling the user to update is clear enough.

```suggestion
If you have HTTPS for all your URLs and still saw this prompt, please update to the latest app version.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+## What do "Most secure" and "Less secure" mean?
+
+- **Most secure:** The app will only use unencrypted connections when you're on your specified home network. This requires location permission so the app can check your current Wi-Fi SSID.
+- **Less secure:** The app will use unencrypted connections regardless of which network you're on. This is not recommended if you ever connect from public Wi-Fi networks.
```
jpelgrom: This also applies to background webhooks/sensor updates so any connection to a public Wi-Fi network is relevant, not just opening the app and connecting (to your server) from public Wi-Fi.

```suggestion
- **Less secure:** The app will use unencrypted connections regardless of which network you're on. This is not recommended if you ever connect to public Wi-Fi networks.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+:::warning
+Choosing "Less secure" may expose your Home Assistant credentials if you connect from untrusted networks.
```
jpelgrom: untrusted network != not home network. For example, I trust the network at work but that is not my home network and still exposes my credentials.

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+## How do I set up my home network?
+
+1. Grant location permission when prompted (required for the "Most secure" option).
+2. Connect to your home Wi-Fi network.
+3. Go to **Settings** > **Companion app** > **[your server]** > **Internal URL**.
+4. Add your home Wi-Fi SSID to the list.
+
+You can add multiple SSIDs if you have more than one home network.
+
+## Can I specify the Hardware Address (Ethernet) instead of just the SSID?
```
jpelgrom, line 62: This sounds like you always need to specify an SSID and cannot only use the MAC address?

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+No. The connection security level feature uses your Wi-Fi SSID to determine if you're at home. When connected via VPN, the app cannot detect your physical location through Wi-Fi.
+
+If you rely on VPN for remote access, you'll need to configure your setup differently:
+
+1. Set your local Home Assistant address as the **external URL** (since you access it the same way whether home or away).
+2. Ensure your VPN is connected whenever you want to interact with Home Assistant remotely.
+
+:::warning
+If you set your local (HTTP) address as the external URL and forget to connect your VPN on a public network, your credentials could be transmitted unencrypted.
```
jpelgrom: I would try to use a similar warning text here to the previous one, since it talks about the same risk.

```suggestion
If you set your local (HTTP) address as the external URL and forget to connect your VPN on a public network, this may expose your Home Assistant credentials.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+:::
+
+## I use VPN for all remote access. Should I enable this feature?
+
+If you've set your local address as the external URL, the connection security level setting won't affect how the app connects—it will always use that external URL.
```
jpelgrom: This may be personal preference, but for clarification without dramatic pause don't use an em dash but a normal colon instead.

```suggestion
## I use VPN for all remote access. Should I enable this feature?

If you've set your local address as the external URL, the connection security level setting won't affect how the app connects: it will always use that external URL.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+However, for advanced use cases, you may still want different internal and external URLs. In that case, enable the "Most secure" option to ensure your internal (possibly HTTP) URL is only used when you're actually at home.
+
+## Will the app send my credentials if Home Assistant is unreachable?
```
jpelgrom, line 85: The question is simple but the answer feels complicated to me while it is actually quite simple (yes - it will attempt to connect which includes your credentials). The difference about encrypted or not should be split into a different question or removed altogether as it is already mentioned a couple of times.

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+The app will attempt to connect using your configured URLs. If the external URL is an HTTPS address, your credentials are encrypted during transmission regardless of whether the connection succeeds.
+
+If your external URL is HTTP (not recommended for remote access), credentials would be sent unencrypted when the app attempts to connect.
+
+## I chose "Less secure" but now want to change it. How?
+
+Go to **Settings** > **Companion App** > **[your server]** > **Connection Security Level** to change your preference at any time. To define your home network, go to **Settings** > **Companion App** > **[your server]** > **Internal URL**.
```
jpelgrom: Home Assistant uses sentence case

```suggestion
Go to **Settings** > **Companion app** > **[your server]** > **Connection security level** to change your preference at any time. To define your home network, go to **Settings** > **Companion app** > **[your server]** > **Internal URL**.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+## What happens if I don't grant location permission?
+
+Without location permission, the app cannot determine which network you're on. If you choose "Most secure" without granting location access, the app will only be able to use your external (HTTPS) URL.
```
jpelgrom:
```suggestion
Without location permission, the app cannot determine which network you're currently connected to. If you choose "Most secure" without granting location access, the app will only be able to use your external (HTTPS) URL.
```

docs/getting_started/connection_security_level.md
```diff
@@ -32,3 +32,68 @@ When you connect to Home Assistant using an unencrypted URL (such as `http://hom
...
+
+If you need to use an unencrypted internal URL and don't want to grant location permission, you'll need to select "Less secure"—but be aware of the security implications.
```
jpelgrom: The text at the end could link back to the FAQ about the difference of most/less secure which mentions the security implication.

docs/getting_started/connection_security_level.md
bgoncal: On iOS that's true

docs/getting_started/connection_security_level.md
bgoncal: The general idea of the warning is that it will expose your credentials and that's critical in networks you do not trust.

How would you prefer to phrase it?

docs/getting_started/connection_security_level.md
jpelgrom: Weird but OK. I would expect to be able to set it separately, what if you only use ethernet?

docs/getting_started/connection_security_level.md
jpelgrom: "... if you connect from a public network. Make sure you trust all networks you connect to when using this option."?

Similar to the other warning it uses "public network", and still mentions trust.

docs/getting_started/connection_security_level.md
bgoncal: I rephrased

PR #1297, opened 2026-03-03
Add documentation for progress bar customization on Android

inukiwi:
## Proposed change
This feature adds support for progress bar customization on Android 16.0+ devices, as shown [here](https://developer.android.com/develop/ui/views/notifications/progress-centric).
Screenshot of an example notification:
image

## Type of change

- [x] Document new or changing features for which there is an existing pull request elsewhere
- [ ] Spelling or grammatical corrections, or rewording for improved clarity

## Checklist

- [x] I have read and followed the [documentation guidelines](https://developers.home-assistant.io/docs/documenting/general-style-guide).
- [x] I have verified that my changes render correctly in the documentation.

## Additional information

- Link to relevant existing code or pull request: https://github.com/home-assistant/android/pull/6527


docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
             progress_max: 32
 ```
 
+#### Styling the progress bar <span class='beta'>BETA</span> {#styling-the-progress-bar}
+
+On Android 16.0+ it is also possible to customize the progress bar with segments, points and icons.
+
+When using any of these options together with a video attachment, the video attachment will not be shown. The `progress_indeterminate` parameter will also be ignored.
```
TimoPtr, line 625: Put this into a :::note

docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
...
+
+The following parameters are supported, they can be mixed and matched together:
+
+- progress_segments - a list of items containing a `length` property and an optional `color` property
```
TimoPtr:
```suggestion
- `progress_segments` - a list of items containing a `length` property and an optional `color` property
```

I think we should make sure to highlight what are the parameters using proper code blocks. Check the rest bellow.

docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
...
+  - When using segments, `progress_max` will be ignored and will instead be the sum of the length of the segments
```
TimoPtr: Could be a note?

docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
...
+- progress_points - a list of items containing a `position` property and an optional `color` property
```
TimoPtr: define the acceptable range from position ? and maybe color too (we cannot write red for instance)?

docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
...
+- progress_(start|end|tracker)_icon - an icon formatted as in Home Assistant for example: `mdi:cellphone`
```
TimoPtr: You can maybe mention that it actually only supports mdi

docs/notifications/basic.md
```diff
@@ -617,6 +617,51 @@ automation:
...
+  - `start` will be shown in front of the progress bar
+  - `end` will be shown at the end of the progress bar
```
TimoPtr: Does it respect RTL/LTR ?

docs/notifications/basic.md
c0ffeeca7:
```suggestion 

##### Progress segments

- `progress_segments`. A list of segments that make up the progress bar, each with its own length and optional color. For example: `[{length: 80}, {length: 20, color: "#ff0000"}]`.
  - **Required properties**: `length`. The length/size of the segment.
  - **Optional properties**: `color`. Hex color code for the segment (for example, `"#ff0000"` or `"#00ff00"`).
  - **Note**: When using segments, `progress_max` is automatically calculated as the sum of all segment lengths and should not be specified separately.
```
What about a list format? Something like this?

docs/notifications/basic.md
c0ffeeca7:
```suggestion
```

docs/notifications/basic.md
c0ffeeca7:
```suggestion

##### Progress points

- `progress_points`: Markers displayed as points along the progress bar to highlight specific milestones or thresholds. For example: `[{position: 80, color: "#ffffff"}]`.
  - **Required properties**: `position`. The position on the progress bar where the point should appear (must be between 0 and `progress_max`).
  - **Optional properties**: `color`. Hex color code for the point (for example, `"#ffde03"`).
```

docs/notifications/basic.md
c0ffeeca7:
```suggestion

##### Progress icons

- `progress_start_icon`: Icon displayed at the beginning (left side) of the progress bar. For example: `"mdi:battery-10"`.
  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:home` or `mdi:download`).

- `progress_end_icon`: Icon displayed at the end (right side) of the progress bar. For example: `"mdi:battery"`
  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:flag-checkered` or `mdi:check-circle`).

- `progress_tracker_icon`. Icon displayed in a circular badge that tracks the current progress position. For example: `"mdi:lightning-bolt"`.
  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:car` or `mdi:walk`).
  - **Note**: The tracker icon has a colored circular background for emphasis.

```

docs/notifications/basic.md
c0ffeeca7:
```suggestion
**Icon colors**

- `progress_start_color`. Color for the start icon. For example: `"#ff0000"`.
  - **Default**: System accent color.

- `progress_end_color`: Color for the end icon. For example: `"#00ff00"`.
  - **Default**: System accent color.

- `progress_tracker_color`. Color for the tracker icon. For example: `"#ffde03"`.
  - **Default**: System primary color.

```

docs/notifications/basic.md
TimoPtr: I would also define the range of `length`

docs/notifications/basic.md
inukiwi: For color: some color names are accepted so 'red' would work, but I guess for simplicity's sake it's better to just mention hex format in the documentation

docs/notifications/basic.md
inukiwi: RTL/LTR is respected, I'll reword "in front" to "at the start" for consistency and clarity

docs/notifications/basic.md
inukiwi: I love it, makes it a lot easier to show the constraints etc.

docs/notifications/basic.md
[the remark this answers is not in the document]
inukiwi: This is already in an Android specific section so not necessary

docs/notifications/basic.md
inukiwi: I think it's better to leave the example for the progress_segments out of this. The JSON formatting could be confusing and there's a YAML example at the bottom of the section

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+            progress_points:
+            - position: 80
+            progress_start_icon: mdi:battery-10
+            progress_end_icon: mdi:battery
+            progress_tracker_icon: mdi:lightning-bolt
+            progress_tracker_color: "#ffde03"
+            tag: progress_style_notification
+```
+
+This example will result in a notification like this:
```
c0ffeeca7:
```suggestion
This example will create a notification like the one shown here:
```
tiny tweak

c0ffeeca7 (approved):
Nice! LGTM.

bgoncal (changes requested):
Our Live Activity and LIve Updates docs will be combine soon in https://github.com/home-assistant/companion.home-assistant/pull/1303

I think we should wait and add these changes there, what do you think?

inukiwi:
This specific feature is not ready to be released yet so I think it's better to just move on with #1303.
I'll update this PR according to the changes in there when this feature is ready.

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+
+- `progress_points`: Markers displayed as points along the progress bar to highlight specific milestones or thresholds.
+  - **Required properties**: `position`. The position on the progress bar where the point should appear (must be between 0 and `progress_max`).
+  - **Optional properties**: `color`. Hex color code for the point (for example, `"#ffde03"`).
+
+##### Progress icons
+
+- `progress_start_icon`: Icon displayed at the start of the progress bar. For example: `"mdi:battery-10"`.
+  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:home` or `mdi:download`).
+- `progress_end_icon`: Icon displayed at the end of the progress bar. For example: `"mdi:battery"`
```
bgoncal, line 647: Do we need to define start and end? Can't we make it generic so the user defines a bunch of points and we just render based on that definition?

bgoncal: Think of the "Point" as an object containing:
- Length
- Icon
- color
- maybe more properties..

bgoncal: Would be nice to have a "accessibility_name" as well for people using `voice over`

inukiwi: We could probably make it more generic, but if the goal is to make it easier for the end user I think it would make things more complicated by explaining all the restrictions.

A point can only have a position and color, it will always be rendered as a small dot on the progress bar and can't have any icon. There can be unlimited points on a progress bar.
The icons are placed next to the progress bar and not on it, so they look very different from points.
The exception to that could be the tracker icon, which **is** placed on the progress bar, but it's still different from a point.

In the image below you can see the difference: the arrow and the car are trackers, the pink square is a point.

image

bgoncal: Understood, but iOS has more flexibility in this situation (for the first time 🙌🏻), when this gets implemented in iOS we could allow users to customize their points, for example:

In a "waiting for package delivery" scenario:

> Point 1:
> - Warehouse icon
> - Color blue
> - length 0 (if I understood correctly this would mean the start)
> - accessibility name: At warehouse
> 

> Point 2:
> - Truck icon
> - Color brown
> - Length 10
> - accessibility name: On it's way

> Point 3:
> - House icon
> - Color green
> - Length 20
> - accessibility name: Delivered


I think this way we keep it more future proof, what do you think? We can have default values as well so the user does not need to fill everything

inukiwi: Could we just add those properties (icon, accessibility name) as optional and iOS-only properties on the points?

Downside is that the start/end/tracker icons would have to be separately implemented on iOS, but they do look significantly different from points so that might be nice to have anyways. In that case we can have a shared basic set of features (as defined by the Android limitations) and iOS can build extra on top of that.

You can see how the start and end icon are shown in the original image I put in this PR.

bgoncal: We could differentiate, but what prevents android to use the same model? You just need to ignore what you dont use, I mean, you do not need to implement it completely now, what is just for iOS can be done later, just implement in a way that when iOS adds new properties it wont crash the android app.

Do you see a challenge having an array of points where android gets the first and the last to define the start and end?

bgoncal: Oh, I missed the tracker... not sure how to fit this one in, thoughts @TimoPtr ?

inukiwi: We can definitely implement logic to work with that approach, I'm just afraid that we'll end up with more complicated documentation. 

If we'd take this approach I think we'd have to use the following rules:

- If a point is located at position 0 **and** has an icon set, it will be displayed in front of the progress bar as a separate icon
- If a point is located at position `progress_max` **and** has an icon set, it will be displayed after the progress bar as a separate icon
- A point that has an icon set and is not at position 0 or `progress_max` will be displayed as a tracker _(this does mean that a tracker can't be at the absolute beginning or end)_
- There can only be one tracker point, if multiple points match those conditions, the first? will be selected.

bgoncal: Depends how you write it, I'd say we explain in a simple manner + examples + one image pointing to the part of the UI the YAML refers to (have a UI in HA instead of yaml would be lovely but... out of scope).

_"To define your notification progress indicators you just need to provide an `array` of points, look at the example below"_

`<yaml example>`

_"The points located at position 0 or at progress_max can define icons to be visible at the beginning and end of the progress bar, the other points do not support icons and will be displayed along the way."_
_"You can define only 1 point-tracker to indicate the progress"_

_"Check how it looks like in a package delivery scenario following the image below"_

`<Mocked notification image with poins overlayed>`

bgoncal: My example above has the assumption we add a key "is_tracker" or something like that

TimoPtr: I would like to see the two proposal face to face to decide. I agree that point seems more generic but at the same the logic to display before and after the progress bar is not friendly

bgoncal: Perhaps thats where we draw the line and we start differentiating, android specific use cases and iOS specific. If the original proposal is implemented I can make it happen in iOS as well and then for more customization a third type can be introduced just for iOS.
I hope google does not make this more flexible in the future, otherwise the original proposal wont be able to accomodate the changes.

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+When using any of these options together with a video attachment, the video attachment will not be shown. The `progress_indeterminate` parameter will also be ignored.
+:::
+
+The following parameters can be mixed and matched together to style the progress bar:
+
+##### Progress segments
+
+- `progress_segments`. A list of segments that make up the progress bar, each with its own length and optional color.
+  - **Required properties**: `length`. The length/size of the segment, any number larger than 0.
+  - **Optional properties**: `color`. Hex color code for the segment (for example, `"#ff0000"` or `"#00ff00"`).
```
jpelgrom: Why does this mention properties (multiple) when both only have one single property?

jpelgrom: Nitpicking: do you need to include the `#`? The color code is only the 6 hexadecimal numbers.

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+  - **Note**: When using segments, `progress_max` is automatically calculated as the sum of all segment lengths and should not be specified separately.
```
jpelgrom: What happens if you specify it separately? If nothing happens, don't mention "and should not be specified separately".

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+
+##### Progress points
+
+- `progress_points`: Markers displayed as points along the progress bar to highlight specific milestones or thresholds.
+  - **Required properties**: `position`. The position on the progress bar where the point should appear (must be between 0 and `progress_max`).
+  - **Optional properties**: `color`. Hex color code for the point (for example, `"#ffde03"`).
```
jpelgrom: Why does this mention properties (multiple) when both only have one single property?

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+
+##### Progress icons
+
+- `progress_start_icon`: Icon displayed at the start of the progress bar. For example: `"mdi:battery-10"`.
+  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:home` or `mdi:download`).
+- `progress_end_icon`: Icon displayed at the end of the progress bar. For example: `"mdi:battery"`
+  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:flag-checkered` or `mdi:check-circle`).
+- `progress_tracker_icon`. Icon displayed in a circular badge that tracks the current progress position. For example: `"mdi:lightning-bolt"`.
+  - **Format**: Material Design Icons only, use format `mdi:icon-name` (for example, `mdi:car` or `mdi:walk`).
+  - **Note**: The tracker icon has a colored circular background for emphasis.
```
jpelgrom: You already mentioned this except for the "colored" part in the line describing "displayed in a circular badge".

```suggestion
```

docs/notifications/basic.md
```diff
@@ -617,6 +617,77 @@ automation:
...
+```yaml
+automation:
+  - alias: Notify of car charging progress
+    trigger:
+      ...
+    action:
+      - action: notify.mobile_app_<your_device_id_here>
+        data:
+          title: "Live update"
+          message: "This will show on the always-on display"
```
jpelgrom: Update this to match the example image / update the example image to also have a title

jpelgrom (commented):
Don't want to distract too much from the ongoing discussion and agree with https://github.com/home-assistant/companion.home-assistant/pull/1297#discussion_r3452227448, but here are some more minor suggestions.

docs/notifications/basic.md
inukiwi: `#` is required, I've included the requirement

docs/notifications/basic.md
inukiwi: I've created a draft of what the documentation would look like in my opinion if we go for a points-only approach: https://github.com/inukiwi/companion.home-assistant/commit/0f88d2772dc99bb51cd1aa049be973f0df965296
It doesn't make it too complex in this form, but it doesn't cover details like the tracker point looking different from other points etc.

PR #1369, opened 2026-07-03
Document iOS snooze notification actions

bgoncal:
## Summary

Documents the built-in snooze actions on iOS and macOS in the actionable notifications page.

The app automatically adds snooze actions to notifications, so users do not need to configure them. This adds a short section explaining that:

- Users get snooze options by default and can add their own durations or turn off the ones they do not want, managed under Settings, Notifications, Snooze Actions.
- When a snoozed notification comes back, its title is prefixed with "↺" so it is clear it was snoozed.

Companion to the app change that adds the "↺" prefix to re-delivered snoozed notifications.

### DIFF
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -99,6 +99,14 @@ action:
             icon: "sfsymbols:bell.slash"
 ```
 
+### ![iOS](/assets/iOS.svg) Snooze actions
+
+On iOS and macOS the app adds snooze actions to notifications automatically, so you do not need to configure them. Snoozing hides the notification and shows it again after the amount of time you pick.
+
+You can manage these under Settings, Notifications, Snooze Actions. Some durations are enabled by default, and you can add your own durations or turn off the ones you do not want.
+
+When a snoozed notification comes back, its title is prefixed with "↺" so you can tell it was snoozed.
+
 ### `uri` values
 
 To navigate to a frontend page, use the format `/lovelace/test` where `test` is replaced by your defined [`path`](https://www.home-assistant.io/dashboards/views/#path) in the defined view. If you plan to use a dashboard the format would be `/lovelace-dashboard/view` where `/lovelace-dashboard/` is replaced by your defined [`dashboard`](https://www.home-assistant.io/dashboards/dashboards) URL and `view` is replaced by the defined [`path`](https://www.home-assistant.io/dashboards/views/#path) within that dashboard. For example:
--- a/docs/notifications/actionable.md
+++ b/docs/notifications/actionable.md
@@ -99,7 +99,7 @@ action:
             icon: "sfsymbols:bell.slash"
 ```
 
-### ![iOS](/assets/iOS.svg) Snooze actions
+### Snooze actions
 
 On iOS and macOS the app adds snooze actions to notifications automatically, so you do not need to configure them. Snoozing hides the notification and shows it again after the amount of time you pick.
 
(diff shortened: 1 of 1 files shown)


docs/notifications/actionable.md
```diff
@@ -99,6 +99,14 @@ action:
             icon: "sfsymbols:bell.slash"
 ```
 
+### ![iOS](/assets/iOS.svg) Snooze actions
+
+On iOS and macOS the app adds snooze actions to notifications automatically, so you do not need to configure them. Snoozing hides the notification and shows it again after the amount of time you pick.
+
+You can manage these under Settings, Notifications, Snooze Actions. Some durations are enabled by default, and you can add your own durations or turn off the ones you do not want.
```
jpelgrom: This is in the companion app settings, right?

```suggestion
You can manage these under **Settings** > **Companion app** > **Notifications** >  **Snooze Actions**. Some durations are enabled by default, and you can add your own durations or turn off the ones you do not want.
```

docs/notifications/actionable.md
```diff
@@ -99,6 +99,14 @@ action:
             icon: "sfsymbols:bell.slash"
 ```
 
+### ![iOS](/assets/iOS.svg) Snooze actions
```
jpelgrom: I think it might be good to drop iOS from here so we can add a tip about snoozing on Android as well, as people might not know about it/ request it on Android after reading this. The paragraph is already clearly labeled iOS/macOS.

```suggestion
### Snooze actions
```

docs/notifications/actionable.md
```diff
@@ -99,6 +99,14 @@ action:
...
+
+On iOS and macOS the app adds snooze actions to notifications automatically, so you do not need to configure them. Snoozing hides the notification and shows it again after the amount of time you pick.
+
+You can manage these under Settings, Notifications, Snooze Actions. Some durations are enabled by default, and you can add your own durations or turn off the ones you do not want.
+
+When a snoozed notification comes back, its title is prefixed with "↺" so you can tell it was snoozed.
+
```
jpelgrom, line 109:
```suggestion

:::tip
Android 8.0+ natively supports snoozing notifications from any app, including Home Assistant. You may need to enable notification snoozing in the system settings for notifications.
:::
```

jpelgrom:
Looks like an iOS release with this feature was made 6 hours ago so no need to wait for release here :)

Edit: or do you want to wait for App Store review to finish @bgoncal?

jpelgrom (approved):

bgoncal:
It's not live in App Store yet, Apple will probably review during the weekend and I'll release on Monday

gbrunow:
This behavior got me by surprise, I've been looking at my scripts thinking Claude had gone rogue at some point.
It feels like this should be an opt-in behavior, not an opt-out. It would be good for it to be configurable via yaml as well.

In my case this was specifically awkward since I have a destructive action which should be at the bottom.

image

I disabled it globally, but it would be better if I could disable it just for specific notifications via yaml.

bgoncal:
@gbrunow Sorry the delay, I don't usually check comments on merged PRs (goes into my notifications filter).

We have a bug on our implementation, those default actions should only be shown if user have not specified actions in Home Assistant already, I will fix that, thanks for reporting.

Regarding the opt-in, I'm on the fence about that, I think the benefit of not having to teach the user how to configure this little help outweighs the discomfort of some users disabling this manually if they dont like. (Of course considering the bugfix above)

gbrunow:
@bgoncal got it, no worries.

Not adding them if actions are included sounds reasonable enough to me.

PR #1378, opened 2026-07-13
Document kiosk screensaver mode and brightness commands

bgoncal:
Documents the two new kiosk screensaver remote commands added in home-assistant/iOS#5069:

- `kiosk_set_screensaver_mode` — sets the screensaver mode (`mode`: `clock`, `dim`, or `blank`).
- `kiosk_set_screensaver_brightness` — sets the screensaver dim level (`level`, a percentage `0`–`100`).

Updates the iOS Kiosk mode page and the Notification commands reference (both command tables + the value note) and adds an example automation.

## Related
- App: home-assistant/iOS#5069
- Relay: home-assistant/mobile-apps-fcm-push#340

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/integrations/ios-kiosk-mode.md
+++ b/docs/integrations/ios-kiosk-mode.md
@@ -101,6 +101,8 @@ The available commands are:
 | --------- | ------ |
 | `kiosk_show_screensaver` | Show the screensaver immediately. Requires screensaver enabled in kiosk settings. |
 | `kiosk_hide_screensaver` | Hide the screensaver and reset the inactivity timer. |
+| `kiosk_set_screensaver_mode` | Set the screensaver mode. Requires `mode`, one of `clock`, `dim`, or `blank`. |
+| `kiosk_set_screensaver_brightness` | Set the screensaver dim level (how bright the screen stays in `dim` mode). Requires `level`, a percentage from `0` to `100`. |
 | `kiosk_show_camera` | Show a full-screen camera stream. Requires an `entity_id` pointing to a `camera.` entity. |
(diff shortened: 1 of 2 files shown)


jpelgrom:
With a screensaver of 'dim', does 100 result in a different brightness than no-screensaver kiosk with `kiosk_set_brightness` to 100? Or are they the same?

bgoncal:
@jpelgrom Good question, I had to manually check to be sure, but yeah it looks the same.
Dim is still relevant though in case the user does not want to rely on push commands

jpelgrom:
OK, does it save the dim value? Or is it a temporary override and could you also use `kiosk_set_brightness` to update the screensaver brightness?

bgoncal:
You configure the screensaver to activate after a specific amount of time passed, and however you defined it to start it will follow, so if you set to start with dim at 100% it will always go back to this when screensaver appears.

The push command updates this setting, after sent it will change the preferences for screensaver dim level.

jpelgrom (approved):
OK then, I was trying to see if it could be combined but doesn't sound like it. Feature might be easier to understand if you also use the app 😅

jpelgrom:
Do you add the waiting for release label or add beta/testflight inline so we can merge? I still don't fully understand iOS prereleases...

bgoncal:
I'm keeping "waiting for release" because the iOS/iPadOS app is not released yet so it can be confuse to the user to see it live, and if I add the beta tag then I have to open another PR to remove so... for small things I prefer this approach.

